PR #69, opened 2026-08-09
feat(unknown-signer-broadcast-seam): broadcastTransaction unknown-signe…

wighawag:
Task: `work/tasks/done/unknown-signer-broadcast-seam.md`

All checks green. Here is the report.

## What I did

Wired the unknown-signer seam into the single broadcast choke point in `packages/rocketh`.

**Drift check first (both premises hold):**
- `broadcastTransaction` (`packages/rocketh/src/environment/index.ts`) is still the single choke point: it is a closure called from exactly two places, `broadcastExecution` and `broadcastDeployment`, and is exported neither from the module nor on the `Environment` interface.
- `account-signability-classification` landed with the assumed shape: `env.addressSignability`, a lowercase-address-keyed view over `Signability = 'local' | 'node' | 'impersonated' | 'unsignable'`, computed after impersonation, returning `'unsignable'` for unseen addresses.

### Production changes

- `packages/rocketh-core/src/types.ts` (additive, pre-authorised by the task): `UnknownSignerPolicy = 'throw' | 'auto'` and `UnknownSignerPolicyFrame = {policy}`; `onUnknownSigner` on `ChainUserConfig`, `ChainConfig`, `ExecutionParams`, and (resolved, required) on `ResolvedExecutionParams.environment`; `pushUnknownSignerPolicy(frame)` / `popUnknownSignerPolicy()` typed on the `Environment` interface so `@rocketh/unknown-signer` can call them with a core-only dependency.
- `packages/rocketh/src/environment/unknownSignerPolicy.ts` (new): `createUnknownSignerPolicyStack(resolvedGlobal)` → `{push, pop, effective}` with `effective() = top?.policy ?? resolvedGlobal`. Module JSDoc documents the dynamic-scope invariant (sequential single-await deploy scripts; `Promise.all` inside one wrapper leaks a frame, per ADR 0006) and that a frame never touches signable accounts.
- `packages/rocketh/src/environment/index.ts`: the seam. Inside the `type: 'object'` branch, `if (env.addressSignability[from] === 'unsignable')` → read the frame stack *inside* that branch → throw `UnknownSignerError({from, to, data, value})` (`from` carried verbatim from the tx, `contract` left unset). The switch over the policy is exhaustive, so adding `'ask'` later is a compile error rather than a silent fallthrough. The old `cannot get signer for ${from}` is replaced by a defensive guard that names the address and the classification when the signability view and `addressSigners` ever disagree.
- Same file: the signer routing is now an exhaustive `switch (signer.type)` naming **all three `Signer` members** — `signerOnly` (local sign then `eth_sendRawTransaction`; what `privateKey`/hardware protocols return), `wallet` and `remote` (both `eth_sendTransaction`). Routing behaviour is unchanged.
- `packages/rocketh/src/executor/index.ts` + `environment/chains.ts`: `onUnknownSigner` threaded exactly like `autoImpersonate` — execution param > chain config > default `'auto'`. `autoImpersonate` itself is untouched.

### Tests (`packages/rocketh/test/`, real environment, local mock provider, no `@rocketh/test-utils`)

`unknown-signer-seam.test.ts` (driving `env.broadcastExecution` and `env.broadcastDeployment` only): story 4 (populated error, node never called), story 5 funnel half (both funnels, deploy has no `to`, value transfer carries `value`), story 6 (mixed run: signable broadcasts before and after the unsignable throw), story 11 (default, explicit `'auto'`, explicit `'throw'` all throw, no prompt/hang); `contract` unset; `type: 'raw'` never reaches the seam; the defensive not-found error (clear, names the address, not a `TypeError`); precedence unit tests plus a chain-level policy reaching a real env; and the **anti-regression**: an `impersonated` account (and a `local` one) still broadcasts with a `'throw'` frame pushed. `unknownSignerPolicy.test.ts` unit-tests the stack plumbing (frame beats global, pop restores, LIFO nesting, pop-in-`finally` on a throwing action, unbalanced pop is a no-op) — explicitly not claimed as a user story.

`pnpm build`, `pnpm typecheck`, `pnpm test` (254 passed), `pnpm format:check` all green. Changeset at `.changeset/unknown-signer-broadcast-seam.md` (`rocketh` minor, `@rocketh/core` minor, `@rocketh/test-utils` patch). `changeset status --since=main` only reports the file as missing because it is still untracked; it reads changesets from `git diff`, so it resolves once the runner commits.

## Decisions (please ratify or reverse)

1. **`ChainConfig.onUnknownSigner` is left optional/undefaulted** while `autoImpersonate` is defaulted to `false` in `chains.ts`. Reason: the `'auto'` default then has exactly one home (`resolveExecutionParams`); defaulting in `chains.ts` too would make "no chain policy" indistinguishable from an explicit chain-level `'auto'` and would outrank any future higher-level default. Recorded as JSDoc at the type.
2. **An unbalanced `pop()` is a no-op, not a throw.** A mis-nested wrapper must not abort a run from inside a `finally`, where it would also mask the error being propagated. Recorded in the module JSDoc; touches the next task (`unknown-signer-package`), which owns the push/pop call sites.
3. **Six tests in `packages/rocketh-test-utils/test/createTestEnvironment.test.ts` were updated** (outside my file fence, test-only). They broadcast from a named bare address without declaring `nodeAccounts`, which is now correctly `unsignable`; they now declare `nodeAccounts`, and the one asserting the old `cannot get signer` message asserts `UnknownSignerError`. Captured durably in `work/notes/observations/harness-tests-broadcast-from-non-node-accounts.md`.

Note for reviewers on user-visible impact: with the default `'auto'`, a named bare-address account that the node does not list in `eth_accounts` and that impersonation did not resolve now throws `UnknownSignerError` instead of failing later at `eth_sendTransaction`. That is the intended behaviour of this spec, and is stated in the changeset.


wighawag:
Approved. Claim-vs-reality: the seam landed where the task said, inside the single broadcastTransaction choke point, branching on addressSignability rather than on signer presence, with the policy frame read INSIDE the unsignable branch. The invariant that bounced the earlier task set (a frame forces throw over ask, never over impersonation) is both structurally encoded and pinned by two explicit anti-regression tests (impersonated and local accounts still broadcast with a throw frame pushed). Cleanup-vs-behaviour: the signer-routing rewrite from if/else to a switch is behaviour-identical, the defensive no-signer-entry guard is kept and tested, the raw already-signed path still returns before the seam, and rocketh-deploy's own guard was correctly left alone for its own task. onUnknownSigner resolves execution param then chain config then auto, mirroring autoImpersonate exactly, with no impersonate value and autoImpersonate untouched. Coherence: the terms match ADR 0006 and the CONTEXT.md signability glossary entry; nothing is re-meant or forked. Destination check: stories 4, 6, 11 and the funnel half of 5 are covered at both public funnels, contract is deliberately unpopulated for the enrichment task, and the typed push/pop pair on the core Environment gives unknown-signer-package exactly the API its task assumes, so merged as written this reaches the spec's goal. Findings are ratification items only: the no-op unbalanced pop, the optional resolved chain-config field, the out-of-fence test-utils edit and its harness fallout, and the Signer enumeration not reaching the done record.

wighawag:

Gate-1 (acceptance, including `test:getting-started`) and Gate-2 both green; this is the third pass, plus an independent fresh-context review that verified behaviourally in a throwaway clone (58 tests in `rocketh`, 18 in `test-utils`, `tsc --noEmit` clean for rocketh/test-utils/deploy).

## Criteria

- **Decides on signability, not signer presence** — branches on `env.addressSignability[from] === 'unsignable'`; `local`/`node`/`impersonated` fall through to unchanged routing.
- **The drift-prone invariant holds structurally AND behaviourally.** The frame stack is read *only after* the unsignable test, with a comment saying why, and two passing tests pin it: an impersonated account with a `{policy:'throw'}` frame pushed still broadcasts, likewise a local one. This is the exact distinction that drifted through three documents and bounced an earlier task set, so seeing it enforced by structure and not just by comment is the thing I most wanted from this diff.
- **`onUnknownSigner: 'throw' | 'auto'`** with execution-param > chain-config > default precedence, tested including the discriminating case (param `auto` beating chain `throw`). No `'impersonate'` value; `autoImpersonate` untouched.
- **Frame is an object**, push/pop helpers typed on the `Environment` interface in `@rocketh/core`, so `@rocketh/unknown-signer` (core-only dep) can reach them.
- **Defensive not-found throw** present and tested: a signable classification with no signer entry raises a clear named error rather than a `TypeError`.
- **Scope fences honoured** — `contract` asserted undefined (enrichment is a separate task), `packages/rocketh-deploy` untouched (its guard is `deploy-unsignable-deployer-reaches-seam`'s), and the `Signer` union enumerated exhaustively with a reference to `CONTEXT.md`.
- **Tests** live in `packages/rocketh/test/`, build a real environment, drive only `broadcastExecution` / `broadcastDeployment`, and import none of `@rocketh/deploy`, `@rocketh/read-execute` or `@rocketh/test-utils` — so the nx cycle stays open.

## The out-of-fence edit, accepted

Six harness tests in `@rocketh/test-utils` were adjusted. That is outside the stated fence, but it is the right call: those tests declared `{deployer: NODE_ACCOUNT}` without telling the mock node it held the account, so under the new seam it is legitimately `unsignable`. The fix declares `nodeAccounts` (what the tests meant all along) and updates the one assertion that pinned the old `cannot get signer` string. No assertion was weakened, and an observation was filed recording the trap for other harness users.

## Non-blocking, tracked

The signer-routing `switch` has no `default`, so a runtime `signer.type` outside the union would fall off the end and return `undefined` as a tx hash, where the old if/else sent it down the sign-then-send-raw path. It needs a user-supplied `signerProtocol` to violate its own type contract to reach, so it is not blocking — but a silent `undefined` at the choke point is the opposite of this feature's purpose, so I am filing it as an observation rather than losing it.

Also noted: the seam-level "honours a pushed frame" test cannot fail while both policy values resolve to `throw`, so it is tautological in this slice (the real precedence is unit-tested on the stack, and its teeth arrive with `'ask'`); and no test discriminates the verbatim-vs-lowercased `from`, since the fixture address has no letters.

PR #70, opened 2026-08-09
feat(unknown-signer-package): @rocketh/unknown-signer package with catc…

wighawag:
Task: `work/tasks/done/unknown-signer-package.md`

Done. Here's what happened and what I changed.

## Context: this was a re-drive, not a fresh build

The branch `work/task-unknown-signer-package` already carried the full prior implementation (commit `7409ee8`). `origin/main` shows exactly why it came back (commits `9521797` + `bd556fe`): Gate‑2 review **blocked on one defect only** — the headline example called `execute(env)` with three positional arguments.

I verified that claim myself rather than trusting it (`packages/rocketh-read-execute/src/index.ts:112-125`): `execute(env)` takes `(deployment, args)` where `args: ExecutionArgs` carries `account`, `functionName` and `args`. I then compiled both forms in a scratch file (deleted afterwards, outside the tracked tree):

- three-arg form → `error TS2554: Expected 2 arguments, but got 3`
- corrected form → compiles clean

## Drift check (the task asked for it first)

Both premises hold on current `main`, so no needs-attention routing:

- policy frame stack: `Environment.pushUnknownSignerPolicy(frame: UnknownSignerPolicyFrame)` / `popUnknownSignerPolicy()` (`packages/rocketh-core/src/types.ts:704-705`), frames are objects carrying `policy` (`:614`), implemented in `packages/rocketh/src/environment/unknownSignerPolicy.ts`.
- `UnknownSignerError` exported from `@rocketh/core` (`packages/rocketh-core/src/index.ts:5`).

## Changes in this pass

- `packages/rocketh-unknown-signer/src/index.ts` — module JSDoc example: correct `execute` call shape, plus the missing `execute` import.
- `packages/rocketh-unknown-signer/README.md` — same fix in the headline snippet.
- `documentation.md` — same fix in the unknown-signer section.
- `README.md` — re-aligned the Packages table via prettier (the new row had left it ragged); table-only diff.

Everything else from the prior attempt was reviewed and kept: the thunk-only divergence (compile error + actionable runtime throw), exact v1 return parity (`toStrictEqual`, keys present when `undefined`, `value` stringified, no `contract`), frame push/pop in `finally`, the ADR‑0006 anti-regression test that an impersonated account still broadcasts inside the wrapper, rethrow-unchanged, cross-realm catch by `name`, and the zero-persistence assertions.

## Decisions (carried over from the prior attempt, recorded durably — linking them here so they are discoverable)

1. **`UnknownSignerError` is re-exported from the `@rocketh/unknown-signer/errors` subpath, not the package root.** Rationale in `packages/rocketh-unknown-signer/src/errors.ts` JSDoc, and the underlying repo-wide constraint in `work/notes/observations/extension-package-root-must-export-only-curried-functions-2026-08-09.md`.
2. **The deferred-tx block prints via `env.showMessage`, not `console.log` as v1 did.** User-visible (respects log level); recorded in the `catchUnknownSigner` JSDoc.
3. **Test-side `@ts-expect-error` assertions are not CI-enforced** because `pnpm typecheck` covers `src/**` only: `work/notes/observations/test-files-are-outside-pnpm-typecheck-2026-08-09.md`. The compile-time half of the thunk divergence was verified manually, as was this pass's snippet fix.

## Checks

`pnpm format:check`, `pnpm typecheck`, `pnpm build`, `pnpm test` (22 files / 276 tests) all green; `changeset status --since=main` reports `@rocketh/unknown-signer` at minor (changeset present at `.changeset/unknown-signer-package.md`). Working tree contains only the four intended file modifications, no untracked scratch.


wighawag:
Approved. Claim-vs-reality: every API this package leans on exists with the assumed shape - pushUnknownSignerPolicy/popUnknownSignerPolicy and showMessage are real Environment members (packages/rocketh-core/src/types.ts:704-726), UnknownSignerError and postfixBigIntReplacer come from @rocketh/core, and the dependency stays core-only as the task required. The defect that blocked the previous round is genuinely fixed: the execute example now uses the real two-argument shape execute(env)(proxy, {account, functionName, args}) in all three places (documentation.md, package README, module JSDoc), matching packages/rocketh-read-execute/src/index.ts. Acceptance conformance: every criterion has a real test - strict-equality return parity with all four keys present, bigint value stringified, no contract on the return, the thunk-only guard at both type and runtime level, push/pop balance including the non-UnknownSignerError path and nesting, the ADR 0006 anti-regression that an impersonated account still broadcasts inside the wrapper, the mixed-run case, print fallback when contract.name is absent, unchanged rethrow, and zero persistence asserted against a spied in-memory store. Coherence: the frame value throw is the existing UnknownSignerPolicy vocabulary, not a fork, and the root export surface stays function-only so the extension spread idiom holds. Destination check: merged as written, stories 2, 3 and 10 land - a v1 script migrates with the import plus the one mechanical arrow-function change, and the divergence fails loudly at compile time and at runtime rather than silently. Five non-blocking ratification findings follow; none of them changes the outcome.

wighawag:

Gate-1 and Gate-2 green after the re-drive; this is the third pass, plus an independent fresh-context review that built and ran the package in a throwaway copy.

## The bounce is genuinely fixed

Gate-2 blocked the first attempt because the headline example called `execute(env)` with three positional arguments. All three copies (module JSDoc, package README, `documentation.md`) now use the real two-argument form with the call args inside the options bag, and a branch-wide grep finds no fourth occurrence. The re-drive changed nothing else of substance.

## Criteria

- **Thunk-only, proven both ways.** The promise form is a compile error (verified as `TS2345`, and the `ts-expect-error` directives were confirmed load-bearing rather than vacuous), and the runtime guard throws an error that *names the fix* rather than failing obscurely. That pairing was the point: a v1 user pasting the old idiom gets told what to change.
- **Strict v1 return parity** via `toStrictEqual` plus an `Object.keys` check: all four keys present even when undefined, `value` stringified, no `contract`. Checked against v1's `helpers.ts`.
- `null` on success; frame pushed before the action and popped in `finally`, including when the action throws a non-`UnknownSignerError`; other errors rethrow unchanged.
- **Runtime dependency is `@rocketh/core` alone**, matching `@rocketh/deploy`; `@rocketh/test-utils` is a devDependency only, so no layering inversion.
- Zero persistence; tests write nothing outside temp fixtures.
- **`covers` stays honest at [2,3,10].** Story 9 is deliberately unclaimed because the frame's precedence cannot be observed while both policy values resolve to `throw`, and the tests do not fake a behavioural difference. A test labelled "story 6" is extra coverage, not a coverage claim.

## Non-blocking, noted

`UnknownSignerError` is re-exported on an `/errors` subpath (justified: `withEnvironment` calls every root export as `func(env)`), but neither the README nor `documentation.md` mentions it, so a consumer wanting an `instanceof` check will not find it. Worth a line when the docs next move.

The deferred-tx block prints through `env.showMessage`, so a run at `--log-level warn` loses output v1 always printed. Defaults print, so it is a documented divergence rather than a miss.

Also: the printed block omits `to:` for a deploy where v1 printed the zero address (the criterion says v1-*style*, not byte-identical), and the nits note records "Gate 2 APPROVED" without mentioning the earlier bounce, so a later reader loses that history.

PR #71, opened 2026-08-09
feat(unknown-signer-contract-enrichment): Populate contract {name?, met…

wighawag:
Task: `work/tasks/done/unknown-signer-contract-enrichment.md`

Done. `pnpm typecheck`, `pnpm test` (285 tests, 23 files) and `pnpm format:check` are all green.

## Drift check (done first, all premises held)

- The seam landed: `broadcastTransaction` in `packages/rocketh/src/environment/index.ts` throws `UnknownSignerError` inside the `addressSignability[from] === 'unsignable'` branch, with `contract` left unpopulated.
- `broadcastExecution` did still call `broadcastTransaction(transaction)` with no options, so the gap was real.
- The legacy `createMockEnvironment` still exists in `@rocketh/test-utils` and carried its own copy of the signature.
- ADR 0006 already says `contract.name` is resolved through the existing `fromAddressToNamedABIOrNull` (no hand-rolled reverse lookup), so no ADR correction was needed.

## What changed

- `packages/rocketh-core/src/types.ts` — `broadcastExecution` options widened to `{message?: string; contract?: Omit<UnknownSignerContractCall, 'name'>}`, with JSDoc on the interface member.
- `packages/rocketh/src/environment/index.ts` — the private `broadcastTransaction` takes an `origin?: {contract?}` parameter (exported surface unchanged: still unexported, still absent from `Environment`); the throw site builds `contract` and resolves `name` via the existing `fromAddressToNamedABIOrNull(to)`, first name when several, absent when none.
- `packages/rocketh-read-execute/src/index.ts` — `execute` (hence `executeByName`) declares `contract: {method, args}`; `tx` is untouched, so plain txs/value transfers stay unenriched.
- `packages/rocketh-test-utils/src/index.ts` — legacy mock mirrors the widened signature (accepts and ignores).
- New test (written before the production code, red 5/8 then green): `packages/rocketh-read-execute/test/unknown-signer-contract.integration.test.ts`, 9 tests driving real `execute`/`executeByName`/`tx`/`broadcastDeployment` through `createTestEnvironment` with an unsignable named account.
- `.changeset/unknown-signer-contract-enrichment.md`.
- Observation captured (out of scope, not fixed): `work/notes/observations/fromaddresstonamedabiornull-can-throw-2026-08-09.md`.

## Decisions

1. **Threading, not catch-and-rethrow** (recorded durably as JSDoc on `broadcastTransaction`, `packages/rocketh/src/environment/index.ts`). The metadata is threaded into the private `broadcastTransaction` so the error is constructed ONCE at the site that already knows the unsignable `from` and owns the message shape; the stack points at the seam. Alternative considered: `broadcastExecution` catching and rethrowing an enriched copy, which keeps the private signature narrow but duplicates error construction (or mutates `error.data`) in every funnel that ever wants to enrich, and creates a second place where the message shape is decided. Touches: `broadcastExecution` only; `broadcastDeployment` passes nothing, so deploys stay unenriched by construction.
2. **Options carry no `name`** — the type is `Omit<UnknownSignerContractCall, 'name'>` rather than a new named concept, so no new vocabulary is introduced and callers cannot supply a competing name; resolution stays at the throw site (matches ADR 0006). Touches: any future caller of `broadcastExecution` that wants enrichment.
3. **A failed name lookup never destroys the error** (code comment at the choice site). `fromAddressToNamedABIOrNull` throws `ABI conflict` when two deployments at one address share a selector, so the lookup is wrapped: the call goes unnamed and the message falls back to `to`. Covered by a test.

Not touched (outside owned files): `@rocketh/deploy`'s `broadcastExecution` calls are factory funding/deploy transactions with no function to name, and `@rocketh/proxy` upgrades go through `execute`, so they are enriched automatically.


wighawag:
Approved. Claim-vs-reality: every premise the task stated holds against the bytes. broadcastExecution really did reach the choke point with no options, the widened option is threaded into the still-unexported broadcastTransaction closure (verified: not on the Environment interface, not in the returned object), and name is resolved through the existing fromAddressToNamedABIOrNull rather than a second reverse lookup, matching ADR 0006 (which already prescribes that helper, so the task's 'note the correction' escape hatch was not needed). Cleanup-vs-behaviour: all eight acceptance criteria are met. The three-package type change is consistent across core, the environment and the legacy test-utils fake; a changeset is present; tests assert the enriched payload from execute and executeByName, the absent-name case, the multi-name case, the ABI-conflict case, and absence from tx, a value transfer and a deploy. Composition: the enrichment lands on the public execute, so @rocketh/proxy and @rocketh/diamond inherit it for free, and the deliberate serialisation against remove-legacy-mock-environment is respected. Coherence: contract reuses the error's own vocabulary via Omit<UnknownSignerContractCall, 'name'>, so nothing forks; only the parameter name origin collides with the existing per-transaction origin (address) in the same file. Destination check: merged as written, a user whose proxy owner is a Safe reads Proxy.upgradeTo(0x...) instead of a bare address, and non-contract paths stay silent, which is exactly the spec's promise. Findings are ratifications and nits, none blocking.

wighawag:

Gate-1 and Gate-2 green; this is the third pass, plus an independent fresh-context review that archived the branch to a scratch copy and re-ran the gates (typecheck green across 18 projects, 285 tests / 23 files, format clean).

## The trap this task was written around

The task warned that passing metadata into `broadcastExecution` is **not** sufficient, because it calls the private `broadcastTransaction` with no options and the throw happens one level down. That is bridged deliberately: the metadata is **threaded** into `broadcastTransaction` via a second private parameter, not caught-and-rethrown, so the error is constructed once at the site that owns `from` and the message shape. The rationale is recorded at the choice site, including why catch-and-rethrow was rejected (a second construction path to keep in sync).

`broadcastDeployment` still calls with no origin, so **deploys are unenriched by construction** rather than by a conditional someone can later break. That is the better shape.

## Criteria

- `name` resolved through the existing `fromAddressToNamedABIOrNull`, not a hand-rolled reverse lookup; several matches take `names[0]` with a note; no match leaves the key **absent** via conditional spread.
- The core `Environment` option type and the legacy `@rocketh/test-utils` mirror moved together, so typecheck stays green workspace-wide.
- `broadcastTransaction` remains unexported and absent from the returned environment.
- Tests live in `packages/rocketh-read-execute/test/`, which is the correct home under the two-homes rule: they need `execute`, so they cannot live in `packages/rocketh/test/` without closing the nx cycle.

## The find worth keeping

Enriching the name uncovered that **`fromAddressToNamedABIOrNull` can throw despite its `OrNull` name** — it merges the ABIs of every deployment at an address and raises on a selector conflict. The call site is guarded, with the right reasoning: enriching a name must never replace the error the user actually needs with an unrelated one, so a conflicting address simply goes unnamed and the message falls back to `to`.

The observation also notes the *other* caller, `read`'s retry path, calls it inside a `catch` where a throw would likewise mask the original error, and correctly leaves that out of scope. That is exactly the capture-don't-expand behaviour I want to see.

## Non-blocking

The new private parameter is called `origin`, which already means the originating *address* of a pending transaction in the same module; `callOrigin` would avoid the second meaning. The absent-name assertions use `toEqual`/`toBeUndefined`, which would also pass for a present `name: undefined` key (no runtime impact). And `options.contract` is a public unvalidated option, so a caller can declare a `method`/`args` pair that does not match `data` — presentation-only, since `to` and `data` still print verbatim.

PR #72, opened 2026-08-09
feat(deploy-unsignable-deployer-reaches-seam): Route deploy's own signe…

wighawag:
Task: `work/tasks/done/deploy-unsignable-deployer-reaches-seam.md`

Done. All acceptance criteria met, working tree clean (only intended changes).

## What I did

**Drift check first (task premises confirmed, with one correction).**
- `packages/rocketh-deploy/src/index.ts` did still do `env.resolveAccount(account)` → `env.addressSigners[address]` → `throw new Error('cannot get signer for ...')` before building the transaction. Confirmed.
- `unknown-signer-broadcast-seam` has landed: the seam is in `packages/rocketh/src/environment/index.ts` (`broadcastTransaction`), keys off `env.addressSignability[from] === 'unsignable'`, and resolves the effective policy as `unknownSignerPolicyStack.effective()` = top frame `?? ` resolved global (`packages/rocketh/src/environment/unknownSignerPolicy.ts`). I matched it by *deleting* the second path, not by adding one.
- Correction worth knowing (not a blocker, fix is identical): the old guard only fired for a deployer passed as a **literal address** (no `addressSigners` entry). A **named** bare-address Safe always has an entry (`{type:'remote', signer: provider}`), so that spelling already reached the seam. Captured in `work/notes/observations/deploy-preguard-never-fired-for-named-safe.md` because ADR 0006's caveat sentence overstates the old blast radius.

**Change** (`packages/rocketh-deploy/src/index.ts`): removed the pre-guard (with a comment at the site explaining why the presence-of-signer question was the wrong one), and removed the `signer: Signer` parameter from `getCreate2Factory` / `getCreate3Factory` plus the now-unused `Signer` type import.

**Tests** (written red first): `packages/rocketh-deploy/test/unknown-signer-deployer.integration.test.ts`, 12 tests on `createTestEnvironment` — unsignable deployer (named and literal spellings, plain and deterministic), nothing saved, policy `'throw'` / `'auto'` / pushed frame, a signable deployer inside a pushed frame still broadcasting, and signable create2/create3 still deploying. The two literal-address tests failed with `Error: cannot get signer for 0x2222…` before the source change.

**Changeset**: `.changeset/deploy-unsignable-deployer-reaches-seam.md` (`@rocketh/deploy: minor`).

## Decisions

- **The `signer` argument to the create2/create3 factory helpers is REMOVED, not kept.** Verified by reading both helpers: neither ever reads `signer`; `getCreate3Factory` only forwarded it to `getCreate2Factory`. Every transaction they send goes through `env.broadcastExecution`, which resolves the signer at the choke point. Alternative considered: keep `env.addressSigners[address]` purely to satisfy the two signatures, which would preserve a lookup with no reader and re-invite the same confusion. Both helpers are module-private (not exported from `src/index.ts`), so this touches no public surface and no other package. Recorded durably as JSDoc above `getCreate2Factory` in `packages/rocketh-deploy/src/index.ts`, and in the changeset.

## Checks

`pnpm typecheck`, `pnpm build`, `pnpm test` (24 files / 297 tests) and `pnpm format:check` all pass; the new test file is prettier-clean too.


wighawag:
Approved.\n\nClaim-vs-reality: the diff does what the task and changeset say. The pre-guard in packages/rocketh-deploy/src/index.ts is gone, replaced by a comment explaining why, and the deploy now funnels through env.broadcastDeployment to the single broadcastTransaction seam, which throws UnknownSignerError for an unsignable from under the effective policy. I independently verified the two load-bearing claims rather than trusting the narrative: the signer argument really was never read inside getCreate2Factory/getCreate3Factory (every tx there goes through env.broadcastExecution), so dropping the module-private parameter is right, and a named bare-address account really does always get an addressSigners entry, so the old guard only ever fired for a literal address.\n\nCleanup-vs-behaviour: nothing is persisted on the error path (the throw precedes savePendingTransaction), matching the no-save test. Signable deploys, including create2 and create3, are untouched and fenced by tests.\n\nCoherence: signability, seam, and the policy frame are used exactly as CONTEXT.md and ADR 0006 pin them; no concept is forked or re-layered. Removing the second guard makes deploy agree with the single-choke-point model instead of running a parallel one.\n\nDestination check: story 5 (the mechanism fires for a deploy) and story 6 (a signable deploy inside a pushed frame still broadcasts) are now both demonstrated in packages/rocketh-deploy/test/. Merged as written, the spec goal is reached.\n\nFour non-blocking findings for ratification, the notable one being that an unsignable deterministic deploy can now broadcast the factory tx, or surface a funding transfer, before the deployment error appears.

wighawag:

Gate-1 and Gate-2 green; plus an independent fresh-context review that proved the behaviour rather than assuming it (with main's `src`, exactly the two literal-address tests fail with `cannot get signer`).

## The shape is right: it DELETES a path rather than adding one

`packages/rocketh-deploy/src/index.ts` now has **zero** references to `UnknownSignerError` and **zero** policy code. The error is still constructed only at the seam inside `broadcastTransaction`, reached via `broadcastDeployment`. No duplicate throw, no parallel policy path — which is exactly what the enrichment task took care to avoid, so the two land consistently.

## The verify-don't-assume instruction was honoured

The task told the builder to check whether the looked-up `signer` was actually read inside the create2/create3 helpers rather than assume it either way. It was genuinely unread (`getCreate3Factory` merely forwarded it), both helpers are module-private with no external caller, so the parameter was **dropped** rather than kept alive by a lookup existing only to feed it. The reasoning sits in a JSDoc at the choice site.

## The find worth keeping

The observation records something that corrects our own ADR: the old pre-guard **never fired for the canonical named-Safe spelling**. A named account declared as a bare address always has an `addressSigners` entry, so only a *literal* address deployer ever died there. ADR 0006 says "deploy performs its own lookup and throws before the choke point", which overstates the old blast radius — and it overstates it for precisely the reason the same ADR gives for why the seam cannot key off signer-entry presence. The fix was needed either way, and both spellings now behave identically, but that is a good catch against a document we wrote.

## Non-blocking

The new test file would not type-check if it were checked: it reads `create2.factory` off `DeterministicDeploymentInfo`, a two-variant union where one variant lacks `create2` and the other has it optional. Invisible today because the package's `tsconfig` include does not cover tests — which is the same gap as `format:check` not covering test files. Worth addressing as its own chore rather than here.

Criterion 3 asked for the parameter decision in the *done record*; the done record is byte-identical to the ready version, so it lives only in the JSDoc and commit body. Both are homes the claim protocol allows, so this is bookkeeping rather than a miss.

PR #73, opened 2026-08-10
feat(unknown-signer-integration-scenarios): Unknown-signer integration…

wighawag:
Task: `work/tasks/done/unknown-signer-integration-scenarios.md`

All acceptance criteria are met, and the repo is green.

## What I did

**Drift check first (as the task demanded).** I verified against current code, not the launch snapshot:
- `catchUnknownSigner(env)(thunk, {log?})` returns `{from, to, value, data}` (all keys present, `value` stringified, no `contract`) and `null` on success: `packages/rocketh-unknown-signer/src/index.ts`.
- The seam lives inside the `unsignable` branch at the broadcast choke point and consults the policy frame there only: `packages/rocketh/src/environment/index.ts:964`.
- Config keys unchanged: `executionParams.autoImpersonate`, `onUnknownSigner: 'throw' | 'auto'`.
No drift, so I proceeded.

**New file: `packages/rocketh-unknown-signer/test/scenarios.integration.test.ts`** (16 tests, each with a JSDoc, each body shaped like a deploy script):
- **Story 1 (headline):** `Vault` behind an ERC173 proxy owned by the Safe. The v2 implementation deploys (signable), and `deployViaProxy`'s upgrade step (`from` = the Safe, read out of the proxy's owner slot) is caught. Asserts the returned `{from, to, value, data}` is exactly `proxy.upgradeTo(newImplementation)`, that the printed block names the sender/target/method/args, that the script continues (a later deploy still runs), and that nothing moved on-chain.
- **Story 5 (transaction-agnostic):** `tx()`, a deploy whose deployer is the Safe (no `to`), an `execute()` (printed block carries `to: 0x… (Registry)` + `method: setTreasury`, while the returned object has no `contract`), and a value transfer (`value: '0xde0b6b3a7640000'`).
- **Story 6 (mixed run):** signable deploy broadcasts and is saved while the Safe call is caught; `eth_sendTransaction` was sent only from the deployer. Plus the same mix inside one wrapper.
- **Story 7 (idempotent re-run):** run 1 defers; the "Safe executes" is modelled by moving the proxy's implementation storage slot; run 2 (fresh environment, same deployment store, `internal.loadDeployments()` as the executor does) skips the upgrade, returns `null`, and sends nothing. A **negative control** re-runs *without* moving the slot and gets the identical deferred tx back, proving the skip comes from chain state. A third test asserts the store's file list is byte-identical before run 1, after run 1 and after run 2, and that no `unsigned/deferred/pending`-ish file ever appears.
- **Story 8 (`autoImpersonate: false`):** unwrapped call throws `UnknownSignerError`; the same call wrapped returns the transaction; and with impersonation left ON the same account is `impersonated` and broadcasts inside the wrapper (ADR 0006, not overridden).

**Supporting changes:** devDependencies `@rocketh/deploy`, `@rocketh/proxy`, `@rocketh/read-execute` (pre-authorised) plus `viem` on `@rocketh/unknown-signer` (runtime dep stays `@rocketh/core` alone); a "Worked examples" pointer in the package README; changeset `.changeset/unknown-signer-integration-scenarios.md` (empty frontmatter, test-only).

## Decisions

1. **`viem` added as a test-only devDependency of `@rocketh/unknown-signer`.** The task pre-authorised the four `@rocketh/*` devDeps but not this. Chosen because the tests assert exact calldata and `encodeFunctionData` is the repo's sanctioned way (`AGENTS.md`: "Use viem for Ethereum utilities"); viem is already a dependency of every package these tests drive. Alternative considered: hard-coding function selectors to avoid the dep, rejected as less legible for a file whose job is documentation. `Abi` is taken from `@rocketh/core/types` so no `abitype` dep was needed. Touches nothing outside this package's devDependencies; the published runtime dependency set is unchanged.
2. **Two `Vault` artifact versions carry hand-written bytecode.** `createMockArtifact` returns identical bytes every call, so a "v2" built from it would be considered already deployed and no upgrade would ever be attempted. The helper `vaultArtifact(1|2)` differs in deployed bytecode (with a valid 2-byte CBOR length suffix so `deploy`'s non-strict matcher actually compares them). Documented at the choice site.
3. **`internal.loadDeployments()` is called in the test harness helper.** `createTestEnvironment` does not reload the store on its own; rocketh's executor does that at the start of every run (`packages/rocketh/src/executor/index.ts:324`), so modelling a re-run requires it. Documented in the helper.

## Observation captured (out of scope, not fixed)

`work/notes/observations/stale-lockfile-importer-my-rocketh-project.md` — my `pnpm install` dropped a stale `my-rocketh-project` importer from `pnpm-lock.yaml`; that directory does not exist and is not a workspace member, so any clean install drops it.

## Verification

- `pnpm test`: 25 files, 313 tests passed (16 new).
- `pnpm typecheck`, `pnpm build`, `pnpm format:check`: green. The new test file also typechecks under strict `tsc --noEmit`.
- `changeset status --since=main` passes once the files are staged (it only reads tracked changeset files; confirmed in a throwaway clone).
- Working tree contains only the intended changes; no scratch files left.


wighawag:
Approved. Claim-check: the new file is test-only, no src changed anywhere; every API it leans on exists with the assumed shape (createTestEnvironment options accounts/nodeAccounts/executionParams/deploymentStore/providerConfig, createMapDeploymentStore, internal.loadDeployments, provider.getRequests/clearRequests, env.addressSignability with the pinned four-state union), and the two storage slots match the EIP173Proxy artifact and the exact slots @rocketh/proxy reads. Acceptance: all five stories land with JSDoc that reads as a script - Safe-governed upgrade caught, printed and non-halting; the four tx-agnostic shapes through one seam with the contract block on the print and absent from the return; a mixed run proving only the deployer broadcast; the execute-on-the-Safe-then-re-run loop with a real negative control (re-run before the Safe acts still defers, so the skip is earned by the storage slot, not by the record); and autoImpersonate false showing both throw and catch plus the impersonated counter-case. Nothing touches disk: the store is Map-backed and the file names are asserted before and after both runs, with no unsigned-tx artefact. Coherence: unsignable, impersonated, the policy frame and the thunk call shape are all used as CONTEXT.md and ADR 0006 define them. Destination check: merged as written, a v1 user reading this file learns exactly how to port a Safe-governed script, so the spec's headline documentation goal is reached. Findings are ratification items and one doc-precision nit, none blocking.

wighawag:

Gate-1 and Gate-2 green, plus an independent fresh-context review on a throwaway clone (25 files / 313 tests, the new file 16/16, typecheck clean, no production source touched).

## These tests discriminate, and that was checked by mutation

The failure mode for a suite like this is passing without proving anything, so the independent review mutation-tested it:

- Deleting the storage write that models the Safe executing kills **exactly** the story-7 skip test, while its sibling negative control still defers. So the idempotent re-run really is chain-derived, not faked. That was the scenario I most expected to be hollow.
- Consulting the policy frame *before* the signability check kills 6 tests, including both mixed-run cases. The invariant that drifted through three documents is now pinned by the integration suite as well as the seam's own tests.
- Blanking the seam's `contract` enrichment kills 2; leaking `contract` into the return kills 5.

## The trap was avoided

Story 5's plain-transaction case uses the `tx()` helper, never a `type: 'raw'` `TransactionToBroadcast` — which could not produce an `UnknownSignerError` at all, since it is pre-signed and returns before any signer lookup. Zero occurrences of the raw form in the file.

## It also fixed a regression none of us caught

This branch's `pnpm install` removed a stale `my-rocketh-project` importer from `pnpm-lock.yaml` and filed an observation about it. Worth stating plainly: that importer was **reintroduced by PR #70**, undoing the fix in `08c3037`, and it slipped past Gate-1, Gate-2, my Gate-3 and an independent review, because none of us diffed the lockfile on that PR.

I have traced the root cause and will file it separately: `test:getting-started` scaffolds `my-rocketh-project/` inside the repo, installs there, then `rm -Rf`s the directory — but the importer it adds to the **root** lockfile survives. So the gate itself re-pollutes the lockfile, and both my earlier fix and this one are symptom fixes. That belongs in its own chore, not here.

## Non-blocking

Story 7's JSDoc overclaims slightly: the *upgrade* skip is chain-derived, but the v2 implementation *deploy* is skipped from the reloaded deployment record, so "entirely from on-chain state" is not quite right for the whole scenario. Three of sixteen `it()` blocks lack their own JSDoc (their `describe` carries one). The story-7 negative control uses `toStrictEqual` between runs, which would pass vacuously if both were null. And `viem` was added as a devDependency beyond the four the task pre-authorised — test-only, pinned repo-wide, runtime dependency still `@rocketh/core` alone, so I am ratifying it here.

PR #74, opened 2026-08-10
feat(migrate-deploy-and-read-tests): Migrate deploy and read-execute te…

wighawag:
Task: `work/tasks/done/migrate-deploy-and-read-tests.md`

Done. Everything is green and the working tree is clean.

`createTestEnvironment` landed **with** every option this migration needs, verified in `packages/rocketh-test-utils/src/test-environment.ts`: named accounts in the real `UserConfig.accounts` shape (`accounts`), `eth_accounts` contents (`nodeAccounts`), impersonation success/failure (`impersonation`) plus `autoImpersonate` via the `executionParams` pass-through, the generic partial `UserConfig`/`ExecutionParams` pass-through (`config` / `executionParams`), the mock-provider handle at value and type level, and a per-transaction `contractAddress` on the default receipt. **No drift** — proceeded.

## What changed

- `packages/rocketh-deploy/test/deploy.integration.test.ts` — swapped to `await createTestEnvironment(...)` (named accounts declared as bare addresses + `nodeAccounts`, so they are node-held and signable). Added a `The Real Broadcast Path` describe with two tests that can only pass through production's `broadcastTransaction`.
- `packages/rocketh-deploy/test/deploy-value.integration.test.ts` — swapped to `createTestEnvironment`; the provider type reference is now `Awaited<ReturnType<typeof createTestEnvironment>>['provider']`. Both `value` assertions are unchanged and now observe the transaction the real choke point dispatched.
- `packages/rocketh-read-execute/test/read.integration.test.ts` — swapped to `createTestEnvironment()` (a read needs no account).
- `.changeset/migrate-deploy-and-read-tests.md` — empty changeset (tests only, no published code changed), in the exact shape `changeset add --empty` produces.

No harness change was needed: **zero provider tweaks, zero stateful chain simulation**. The existing canned defaults answered every RPC the real path makes, including the create2/create3 factory bootstrap (`eth_getCode` → `0x`, funded deployer balance, raw factory deploy).

## The broadcast-path proof

`deploy.integration.test.ts` → `The Real Broadcast Path`:
1. **signer selection follows `addressSigners`**: a deployer declared with a private key resolves through the shipped `@rocketh/signer` `privateKey` protocol to `signerOnly`, so the deploy is signed locally and the node only ever sees `eth_sendRawTransaction` — asserted together with `expect(methods).not.toContain('eth_sendTransaction')`. The legacy fake always called `eth_sendTransaction` for an object transaction regardless of signer, so this cannot pass against it.
2. **`evm_mine` on `autoMine`**: asserted to land *after* the send and *before* the receipt poll, which is where production emits it inside `broadcastTransaction`.

## Assertion changes (the two places the old fake's shortcut was load-bearing)

1. `should demonstrate idempotent deployment with skipIfAlreadyDeployed` — the stale comment "In mock, newlyDeployed is always true, but pattern is demonstrated" was an (incorrect) claim about the fake, and the test only asserted `toBeDefined()`. Against the real environment the second call genuinely reuses the saved deployment, so I **added** `newlyDeployed === false` and same-address assertions. Nothing was removed or weakened.
2. `should use custom retry config from environment` — previously poked `(env.context as any).retry = ...` after construction, which only worked because the fake's `context` was a fabricated literal. It now sets the same budget the way a user does, through `config: {retry: {maxRetries: 2, delay: 100}}`, so `env.context.retry` is the value production resolved. The `callCount === 3` assertion is unchanged.

Everything else is a mechanical `createMockEnvironment()` → `await createTestEnvironment(...)` swap. No test deleted, skipped or weakened. `createMockEnvironment` is untouched and still used by `rocketh-proxy`/`rocketh-diamond` (batch 2) and the harness's own test. I did not touch the two batch-2 files.

## Observations captured (out of scope, not fixed)

- `work/notes/observations/multi-account-deploy-test-uses-one-account-2026-08-10.md` — the pre-existing `deploying with different named accounts` test passes `user1` for both deploys, so `user2` is never exercised. Preserved as-is.
- `work/notes/observations/format-check-covers-src-only-2026-08-10.md` — `format:check` globs `src/**` only; three *other* test files are already unformatted.
- The `pnpm test:getting-started` lockfile pollution bit me while running the gate locally (it re-added a `my-rocketh-project` importer to `pnpm-lock.yaml`); already captured in `getting-started-test-pollutes-root-lockfile-2026-08-10.md`, and I restored the lockfile so the tree is clean.

## Gate status

`pnpm typecheck` ✅ · `pnpm build` ✅ · `pnpm test` ✅ 25 files / **315 tests** (was 313; +2 new broadcast-path tests) · `pnpm format:check` ✅ · `pnpm test:getting-started` ✅. `changeset status --since=main` still reports "no changesets" only because the new changeset file is untracked; it becomes visible once the runner commits (`git diff` against the merge base cannot see untracked files).


wighawag:
Approved. Claim-vs-reality: the two new broadcast-path tests assert things only a real environment can show, and both check out against production. Signer routing does sign locally then send raw for signerOnly and emits evm_mine after the send (packages/rocketh/src/environment/index.ts:1039-1069), and a bare 66-char key really does resolve through the privateKey protocol to signerOnly (index.ts:289-296). Cleanup-vs-behaviour: all three files are fully off createMockEnvironment, no test deleted, skipped, or weakened; assertions changed in exactly two places and both got stronger (real skip-if-already-deployed, retry budget from resolved config instead of an env.context poke). No harness change at all in this commit, so the no-stateful-chain-simulation fence holds, and the legacy fake is untouched and still used by proxy/diamond. Two pre-existing oddities (a multi-account test using one account, format:check globbing src only) were recorded as observations rather than fixed, which is the right scope discipline. Coherence: signerOnly / test environment / unsignable are used exactly as CONTEXT.md pins them. Destination check: batch 1 of expand-migrate-contract lands complete, leaves batch 2 file-orthogonal, and satisfies remove-legacy-mock-environment's no-straggler precondition for these files. Non-blocking: a config snippet in the new private-key test omits signerProtocols and would fail if copied, and the changed assertions were never noted in a done record as the task asked.

wighawag:

Reviewed the diff against the task's acceptance criteria, independently verifying the load-bearing claims rather than taking the PR body's word for them.

## Per-criterion

1. **All three files use `createTestEnvironment`; no `createMockEnvironment` in them** — ✅ confirmed by grep over the branch. Exactly the three named files changed; the two batch-2 files untouched.

2. **Intent preserved, no test deleted/skipped/weakened; assertion changes noted** — ✅ on substance. Both changed assertions are *strengthenings*, and I ratify them:
   - idempotency test: was `toBeDefined()` plus a stale comment claiming the fake always reports `newlyDeployed: true`; now asserts `newlyDeployed === false` and a stable address, which is the behaviour the criterion was always about.
   - retry test: dropped `(env.context as any).retry = ...` in favour of `config: {retry: {...}}`. This removes an `any` cast and makes the budget the one production resolved. `callCount === 3` unchanged.

   On the *bookkeeping* half, Gate 2's nit is right: the done task file landed unchanged and the commit body is empty, so neither carried the note. I am closing that at the merge rather than bouncing the task — this PR body is being squash-merged as the commit body, so both changes are recorded durably in git history, which is what the criterion exists for.

3. **At least one deploy test proves the real broadcast path** — ✅ delivered twice, and I checked that the first one is genuinely *discriminating* rather than merely present. The legacy fake picks its RPC from `transaction.type === 'raw' ? eth_sendRawTransaction : eth_sendTransaction` (`rocketh-test-utils/src/index.ts:547,566`) and hardcodes every `addressSigners` entry to `type: 'remote'`, so a `signerOnly` deployer is unrepresentable there and the fake would have sent `eth_sendTransaction`. The new test asserting `eth_sendRawTransaction` present / `eth_sendTransaction` absent therefore cannot pass against the old harness. The `evm_mine` ordering test pins the `autoMine` follow-up inside `broadcastTransaction`.

4. **No stateful chain simulation added** — ✅ stronger than required: `packages/rocketh-test-utils/` is not in the diff at all. Zero harness change, so the harness's non-goal was not touched and batch 2 inherits a free hand there.

5. **`createMockEnvironment` still exists and other tests still use it** — ✅ untouched; proxy, diamond and the harness's own test still reference it.

6. **Workspace gate green** — ✅ trusting Gate 1, which under `freshWorktreeGate` ran `prepare` + `verify` in a clean worktree cut from this branch rebased onto `main` (the log shows the getting-started leg executing under the fresh-gate tree), plus Gate 2's approval. Not re-run here by design.

## Lockfile (this repo's standing trap)

Diffed it explicitly, because a phantom `my-rocketh-project` importer has been reintroduced once already by a PR that four review layers missed: `git diff origin/main...HEAD -- pnpm-lock.yaml` is **zero lines**, and `my-rocketh-project` appears **zero** times in the branch's lockfile. Clean. The author hit the pollution while running the gate locally and restored it, which matches the root cause filed in `getting-started-test-pollutes-root-lockfile-2026-08-10.md`.

## Nit triage

- **Worth promoting to a task** — the new private-key test's "Usage in real scenario" snippet shows `accounts.deployer = 'privateKey:0x...'` without `signerProtocols: {privateKey}`. I verified this at source and it is a real documentation defect in new code, in a repo whose convention is that integration tests *are* documentation: `environment/index.ts:290` and `:309` read `userConfig.signerProtocols?.[...]` with nothing registering it by default, so the snippet's `privateKey:` form throws `protocol: privateKey is not supported`, and the bare-66-char form the test itself uses takes an `if (privateKeyProtocol)` that is simply skipped, leaving the account **silently** unresolved. Not blocking (a comment, no behaviour, no criterion), and it has a durable home in the nits file.
- **Cheap follow-ups, correctly non-blocking** — adding `not.toContain('eth_signTransaction')` to prove signing never reached the node; and the create2/create3 tests still only asserting `toBeDefined()` (a pre-existing weakness a harness swap was right to leave alone).
- Both new observations (`multi-account-deploy-test-uses-one-account`, `format-check-covers-src-only`) are legitimate off-path captures, not scope creep.

Merging.

PR #75, opened 2026-08-10
feat(migrate-proxy-diamond-tests): Migrate proxy and diamond tests onto…

wighawag:
Task: `work/tasks/done/migrate-proxy-diamond-tests.md`

Done. This was a continuation, so I only addressed the Gate-3 block and left everything the requeue note verified as good untouched.

## What I did

1. **Verified the block's premise myself before acting** (not taken on the requeue's word):
   - `packages/rocketh-diamond/src/index.ts:124` — `deterministicFacet` defaults to `true` for every facet.
   - `packages/rocketh-deploy/src/index.ts:438` + `:483` — a deterministic deploy computes `expectedAddress` from the create2 factory and passes it to the environment.
   - `packages/rocketh/src/environment/index.ts:846` — `pendingDeployment.expectedAddress || receipt.contractAddress`, i.e. the computed address wins over the receipt.
   - `packages/rocketh-diamond/src/index.ts:435-448` — the diamond proxy is deployed with `deterministic: options.deterministicSalt`, which is `undefined` in that test, so the proxy address really is receipt-derived.
   
   So the old comment's claim (that the block guards against a single-address receipt) was indeed false for the facets.

2. **Reworded the comment only** at `packages/rocketh-diamond/test/diamond.integration.test.ts:114-121`. It now states what the block actually pins: four distinct CREATE2 facet contracts (address computed from each facet's own bytecode and preferred over the receipt), plus the diamond as a separate, receipt-derived, non-deterministic proxy in front of them, and that this is a graph-shape assertion replacing a `toBeDefined()`-only case rather than a receipt guard. **Every assertion kept; nothing weakened.** The proxy test's comment at `:84-95` was left alone (verified accurate: implementation and proxy there are non-deterministic).

3. **Recorded the added-assertions decision durably** at `work/notes/observations/decision-added-assertions-migrate-proxy-diamond-tests-2026-08-10.md` — what was added, why, that the proxy block sits inside acceptance criterion 3's fence while the diamond facet block sits outside it, and the alternative (verbatim port, leaving both cases `toBeDefined()`-only). I cannot write the commit body, so for the runner's commit message:

---
Decisions: two assertion blocks were ADDED rather than only ported, as deliberate strengthenings of `toBeDefined()`-only cases. The proxy block (implementation != proxy, deployment == proxy, merged ABI larger) sits inside acceptance criterion 3's fence: under the old single-`contractAddress` receipt those two collapsed onto one address, so it could not have existed. The diamond facet block (four distinct create2 facets, diamond a separate receipt-derived proxy) sits OUTSIDE that fence, since facets default to `deterministic: true` and their create2 address already beat the receipt. See work/notes/observations/decision-added-assertions-migrate-proxy-diamond-tests-2026-08-10.md for the rationale and the trim-back alternative.
---

## Checks

- `pnpm build` — 17/17 projects OK
- `pnpm typecheck` — clean
- `pnpm format:check` (plus prettier on both touched test files) — clean
- `pnpm test` — 25 files / 315 tests passed
- No `createMockEnvironment` reference remains in either package; no `eth_call` / `eth_getStorageAt` mocks in either suite (fresh path still driven)
- Working tree: only `packages/rocketh-diamond/test/diamond.integration.test.ts` modified and the one new note; lockfile clean


wighawag:
Approved. Both suites are genuinely off the fabricated harness: no createMockEnvironment reference remains in either file (including the commented-out transparent-proxy TODOs), and a repo-wide search shows no test outside test-utils' own deliberate export check still uses it, so the contract step (remove-legacy-mock-environment) is unblocked. Claim-vs-reality holds on the substance: createTestEnvironment exists with the accounts / nodeAccounts shape used, the per-transaction receipt address is real, the deterministic-facet create2 reasoning in the headers matches @rocketh/deploy, and the env.get names asserted (OwnableContract_Implementation / _Proxy, MyDiamond_DiamondProxy, the three _DefaultDiamond* facets) match the sources. No harness change was needed and none was made, so the no-ambient-chain-state fence is respected; no upgrade-path test was added; nothing was deleted, skipped, or weakened into a tautology, and the one real oddity found (three example facets sharing a create2 address) was recorded as an observation rather than papered over. Terminology matches the CONTEXT.md test-environment vs mock-environment glossary entry, and the createEnv / NAMED_ACCOUNTS shape mirrors batch 1, so the two batches read as one convention. Destination check passes: merged as written, the migrate step is complete and every proxy and diamond deployment goes through the real broadcast choke point. Three non-blocking items: one inaccurate inline rationale in the diamond test (its facet assertions are not the receipt-collapse guard the comment claims), plus two in-scope choices to ratify (added assertions beyond the stated fence, and leaving the example-artifact collision as an observation with no follow-up task).

wighawag:

The migration itself is good work and almost everything checks out. I am blocking on ONE thing: a new inline comment states a causal mechanism that is **false**, and I verified it false at three source sites rather than taking Gate 2's word for it.

## The blocking finding

`packages/rocketh-diamond/test/diamond.integration.test.ts:114-127`, the rationale on the new facet-address block:

> "a harness reporting one contract address for every transaction would silently collapse every facet of the diamond onto a single address, and a `toBeDefined()`-only suite would stay green."

That is not what these assertions pin. Facets are **deterministic**, so their addresses are create2-computed and never come from the receipt:

- `packages/rocketh-diamond/src/index.ts:124` — `let deterministicFacet: \`0x${string}\` | boolean = true;` inside the `for (const facet of facetsSet)` loop, so a facet defaults to deterministic.
- `packages/rocketh-deploy/src/index.ts:438` — `expectedAddress = create.getExpectedAddress({salt, bytecode})`.
- `packages/rocketh/src/environment/index.ts:846` — `const contractAddress = pendingDeployment.expectedAddress || receipt.contractAddress;` — `expectedAddress` **wins**.

So under the OLD single-address receipt those four facet addresses would still have been four distinct create2 addresses, and this block would have been green. It is not the receipt guard it claims to be. (The equivalent comment in the proxy test at `:84-95` **is** accurate, and I am not asking you to touch it: implementation and proxy there are non-deterministic and genuinely receipt-derived, so that one really would have collapsed.)

Two reasons this is blocking rather than a nit, even though it is only a comment:

1. Integration tests are documentation in this repo by explicit convention (`AGENTS.md`), so this text is a published claim about how the system works, sitting right next to the assertions it mis-explains — and it contradicts this same file's own header, which correctly says facets go "through the create2 factory, since a facet defaults to `deterministic: true`".
2. `AGENTS.md` carries a standing rule about exactly this failure mode, because the repo has repeatedly lost review cycles to a plausible, secondhand, wrong assertion that became indistinguishable from fact once repeated. A false mechanism committed into a test comment is the seed of the next one.

## The fix (small, and the assertions themselves stay)

Reword the comment to say what the block actually pins: the four facets are four **distinct create2** contracts, and the diamond itself is a **separate, receipt-derived** proxy in front of them (which is the part the per-transaction receipt does matter for, and is what `expect(facetAddresses).not.toContain(deployment.address)` plus the `MyDiamond_DiamondProxy` identity check are really worth). Do not weaken or remove any assertion.

While you are in there, please also record the two ADDED assertion blocks (diamond facet-distinctness, proxy implementation-vs-proxy) explicitly in the commit body. Acceptance criterion 3 fences assertion changes to "those the old fake's shortcuts made necessary"; the proxy block meets that fence, the diamond block does not, and it is fine to keep it as a deliberate strengthening of a `toBeDefined()`-only case — but it needs to be stated as such rather than justified by a mechanism that does not hold.

## Everything else verified and good — do not redo any of it

- **Nothing parked on the legacy harness.** No `createMockEnvironment` reference remains in `packages/rocketh-proxy` or `packages/rocketh-diamond`. This was the hard line and it is met.
- **Same code paths still driven.** Confirmed the fresh-deployment branch is what every case takes: `proxy/src/index.ts:157` and `diamond/src/index.ts:58` branch on `env.getOrNull(name)`, and your recorded-traffic check (no `eth_call`, no `eth_getStorageAt` anywhere in either suite) is the right discriminating evidence. No speculative mocks, no upgrade tests added.
- **Zero harness change**, so the stated non-goal on chain-state simulation was never approached.
- **The commented-out transparent-proxy cases** are still commented, updated in place, neither revived nor deleted.
- **The create2 fixture collision** was the right call: I verified `createExampleArtifact` (`rocketh-test-utils/src/index.ts:297`) overwrites only `.abi` and leaves `createMockArtifact`'s bytecode identical, so three same-bytecode facets at the same salt genuinely share one address. Pre-existing, not migration-induced, correctly captured as an observation instead of fixed inside a migration batch.
- **Lockfile clean.** Diffed explicitly, since a phantom `my-rocketh-project` importer has slipped past four review layers here before: zero diff lines against `main`, zero occurrences in the branch's lockfile.
- Empty changeset is the right shape (tests only, nothing published changed).

wighawag:
Approved.\nClaim-vs-reality: every claim in the two new header blocks and in both inline rationales checks out against the bytes. Facets do default to deterministic (rocketh-diamond/src/index.ts:124), the environment does prefer expectedAddress over receipt.contractAddress (rocketh/src/environment/index.ts:846), the diamond proxy is deployed with deterministic: options.deterministicSalt which is undefined here, so receipt-derived, and the proxy upgrade branch's eth_getStorageAt reads (rocketh-proxy/src/index.ts:441,464) sit behind the prior-deployment check and stay unreached. The Gate-3 block is discharged: the diamond comment now states what the block really pins (four distinct create2 facets plus a separate receipt-derived proxy) and no longer claims to guard a single-address receipt.\nCleanup-vs-behaviour: acceptance criteria all met. Zero harness change (the diff touches only the two test files, a changeset, three notes and the task move), no legacy createMockEnvironment reference remains anywhere outside test-utils' own export test, the commented-out transparent cases stay commented, no test deleted or weakened, and every case still takes the fresh path.\nCoherence: signable / nodeAccounts / test environment are used exactly as CONTEXT.md pins them, and the account fixture mirrors batch 1 rather than forking a new shape.\nDestination check: with batch 1 already merged, this completes the migration off the fabricated harness for every package test file, which is the sequence's goal. Findings are ratifications and nits only.

wighawag:
APPROVE ✅ — Gate-3 re-review (my earlier BLOCK is resolved)

## The blocking finding is fixed, correctly

The diamond comment now says what the block actually pins, and — importantly — explicitly disclaims the mechanism it previously asserted: "this is a shape assertion about the deployment graph the fresh path builds, **not a guard against a single-address receipt**". It also names the reason (facet defaults to `deterministic: true`, create2 address preferred over the receipt). That is accurate, and it no longer contradicts the file's own header.

Checked the surrounding requirements too:

- **No assertion was weakened to make the reword easy.** The assertion set is byte-identical to the blocked revision: 4 facets, 4 distinct, each of the four named deployments present, diamond address identical to `MyDiamond_DiamondProxy`, diamond not among the facets.
- **The proxy comment was left alone**, as asked. I re-verified independently that it is genuinely accurate rather than trusting the nit: in `@rocketh/proxy` the implementation takes `deterministic: options.deterministic || options.deterministicImplementation` (`src/index.ts:138`) and the proxy takes `options?.deterministic` (`:313`) — both unset in the basic ERC173 case, so both really are receipt-derived and really would have collapsed onto one address under the old fake.

## The two added assertion blocks: RATIFIED

`work/notes/observations/decision-added-assertions-migrate-proxy-diamond-tests-2026-08-10.md` states the fence position honestly, including that the diamond block sits **outside** acceptance criterion 3's "only what the old fake's shortcuts made necessary" fence. I ratify keeping both rather than trimming: each replaces a `toBeDefined()`-only case that could not have detected a collapsed deployment graph, and the diamond block is now labelled as the deliberate strengthening it is instead of being justified by a mechanism that does not hold. That was the whole point of the bounce, and it landed.

On the commit-body ask: the record went into a committed note with a `## Decisions` block instead of the commit body. I accept that as the home — it is more discoverable than a commit body — and I am closing the git-history half by squash-merging this PR body as the commit message.

## Re-verified for this revision (not carried over from the first pass)

- **Lockfile:** zero diff lines against `main`, zero `my-rocketh-project` occurrences in the branch's lockfile. Re-checked on the force-updated tip, since this repo has had a phantom importer slip past four review layers.
- **Diff scope:** the two target test files, the changeset, three notes, the done-record rename. No harness change (`packages/rocketh-test-utils/` absent from the diff), so the chain-state-simulation non-goal was never approached.
- **Nothing parked on the legacy harness** — the hard line, and it holds: no `createMockEnvironment` reference remains in either package.
- **Same paths still driven:** fresh-deployment branch everywhere (`proxy/src/index.ts:157`, `diamond/src/index.ts:58` branch on `env.getOrNull(name)`), evidenced by no `eth_call` and no `eth_getStorageAt` in either suite's recorded traffic. No speculative mocks, no upgrade tests.
- **Commented-out transparent-proxy cases** still commented, updated in place, neither revived nor deleted.
- Gate 1 green under `freshWorktreeGate` (clean worktree cut from this branch rebased onto `main`), Gate 2 approved.

## Remaining nits — none blocking, two worth a human's eye

- **Bucket polarity of the new `decision-*` note.** Fair point and I am not resolving it unilaterally: `CONTEXT.md` pins `notes/observations/` as *spotted, unverified, append-only* and puts "what WE decided and why" in `docs/adr/`, and this is the first `decision-*` file in the tree. The substance is right and an ADR would be disproportionate for "we added two test assertions", so I am letting it land and flagging the taxonomy question for the maintainer rather than spending a build cycle on filing.
- **The `NAMED_ACCOUNTS` / `NODE_ACCOUNTS` / `createEnv()` triple is now duplicated in a third test file.** Correctly deferred (the task fenced harness changes), and the obvious generalisation is a shared preset in `@rocketh/test-utils` once the legacy harness is gone.
- The create2 fixture collision still has no follow-up task; nothing false is asserted (that case is `toBeDefined()`-only), so it is a triage item.

Merging.

PR #76, opened 2026-08-10
feat(remove-legacy-mock-environment): Remove the legacy createMockEnvir…

wighawag:
Task: `work/tasks/done/remove-legacy-mock-environment.md`

Done. Contract step completed: the fabricated harness is gone, one builder remains.

**Precondition check (drift check first, as instructed).** Searched the whole repo for `createMockEnvironment` before touching anything. No test used it as a harness: the only test reference was `packages/rocketh-test-utils/test/createTestEnvironment.test.ts:52`, a transitional assertion from the expand step (`'createMockEnvironment is still exported (this task adds; it does not migrate)'`), i.e. it asserted the export existed rather than building an environment with it. Both migration batches (`@rocketh/deploy`, `@rocketh/read-execute`, `@rocketh/proxy`, `@rocketh/diamond`) had fully moved to `createTestEnvironment`. So no straggler, no needs-attention route. One small drift, harmless: the task said `AGENTS.md` names the old harness "in both its `Do` list and its test-structure example"; the example had already been migrated, only the `Do` bullet named it.

**Changes**

- `packages/rocketh-test-utils/src/index.ts`: deleted `createMockEnvironment`, its `DEFAULT_NAMED_ACCOUNTS`, and its `MockEnvironmentOptions` / `MockEnvironmentResult` types (306 lines), including its own `broadcastExecution` / `broadcastDeployment` reimplementations. Now-unused type imports (`Environment`, `Deployment`, `PartialDeployment`, `TransactionToBroadcast`, `UnknownSignerContractCall`, `EIP1193TransactionReceipt`) dropped. Kept: `createMockProvider` + `MockProvider`/`MockProviderConfig` (used by the real harness), `createMockArtifact`, `createMockArtifactWithLibrary`, `createExampleArtifact`, the `Environment` type re-export, and the `createTestEnvironment` re-exports.
- `packages/rocketh-test-utils/test/createTestEnvironment.test.ts`: test written first, red before the deletion. The transitional assertion was **inverted, not deleted**: it now asserts the package exports exactly one `create*Environment` builder (`['createTestEnvironment']`), a regrowth fence that catches a re-forked fabricated environment under any name. Verified red by temporarily re-adding a `createMockEnvironment` export (`expected [ 'createMockEnvironment', …(1) ] to deeply equal [ 'createTestEnvironment' ]`), then restored. No test deleted, skipped, or weakened; suite count unchanged at 315.
[...]

PR #78, opened 2026-08-10
feat(prompt-capability-on-the-environment): Carry a text-prompt capabil…

wighawag:
Task: `work/tasks/done/prompt-capability-on-the-environment.md`

Both requeue items addressed; everything else from the prior attempt left intact.

## What I did (continuation)

Reviewed `git diff origin/main...work/task-prompt-capability-on-the-environment` first: the prior attempt's implementation (additive `promptText`, `ExecutionParams.promptExecutor`, `env.canPromptForText()`, node impl keyed off `request.name`, web untouched, tests, changeset, ADR ref) is intact and I did not restart or churn any of it. I applied only the two Gate-3 BLOCK fixes, both doc-only, no behaviour change:

1. **`packages/rocketh-core/src/types.ts`** — `TextPromptAnswer` JSDoc no longer lists "an empty non-answer" under what `cancelled` covers. It now states positively that this is a generic text primitive: an empty string is a **value** (`{value: ''}`), and the caller validates (a hash resolver must reject `''` itself and decide re-ask/abort/defer). Now agrees with `createNodePromptExecutor().promptText` (cancelled only when `typeof value !== 'string'`) and with the test that pins `{value: ''}`.
   - Mirrored as a comment at the choice site: `packages/rocketh-node/src/environment/prompt.ts`.
   - Recorded as **decision 6** (what I chose, why, the alternative of mapping `''` to `cancelled`, and what it touches: `ask-policy-interactive-resolver`, `per-call-ask-override-and-deferral-precedence`) in `work/notes/observations/decisions-prompt-capability-on-the-environment-2026-08-10.md`, the same durable home the prior attempt's decisions 1-5 use.

2. **`packages/rocketh/test/prompt-capability.test.ts`** — moved the hardhat-deploy / `loadEnvironmentFromStore` JSDoc block down to sit above the `loadEnvironmentFromStore` test it actually describes; the executor test keeps its own block.

Nothing else touched: no policy branches on the capability, `@rocketh/web` still untouched, the pre-existing confirm `.proceed` bug still only captured as an observation.

## Checks

- `pnpm build` — green (17 projects)
- `pnpm typecheck` — green
- `pnpm test` — green (28 files, 331 tests)
- `pnpm format:check` — green
- Working tree: only the 4 intended modified files, no untracked scratch. Changeset (`@rocketh/core`, `rocketh`, `@rocketh/node`) from the prior attempt already covers the packages I touched.


wighawag:
Approved. Claim-vs-reality holds: the abstraction is widened purely additively (optional promptText, absence is the signal), the prompt rides ExecutionParams and survives resolveExecutionParams by identity, and I independently enumerated the createEnvironment call sites: exactly three (rocketh executor 324 and 422, test-utils 340), no fourth. All three carry the capability, including the hardhat path, which @rocketh/node now feeds from its loader. Node reads the answer keyed by request.name with a regression test using txHash, and Ctrl-C maps to cancelled. Cleanup-vs-behaviour: the inline node prompt moved to createNodePromptExecutor with identical confirm behaviour; the confirm .proceed bug was correctly left alone and captured as an observation; web is untouched and still confirm-only, so the per-capability check reports false for it. Nothing branches on the capability, and a test pins that auto still throws UnknownSignerError without consulting the prompt. Coherence: promptExecutor, canPromptForText and capability-as-ceiling match ADR 0007 and do not re-mean signer or signability; capability sits on the environment, the layer the seam reads, not on the executor. Destination check: merged as written, the enabling slice the spec needs is present on every construction path, and the later resolver tasks can branch on one honest predicate. Findings are ratification items plus a doc/impl mismatch on the empty answer, none blocking.

wighawag:

Strong work, and every acceptance criterion is genuinely met. I am blocking on ONE thing, the same class of defect I bounced `migrate-proxy-diamond-tests` for: a doc claim that contradicts the code, sitting where the next agent will read it.

## Blocking: `TextPromptAnswer`'s doc contradicts its only implementation

`packages/rocketh-core/src/types.ts` documents the type as:

> the fact that they aborted (Ctrl-C, **an empty non-answer**, a runtime that gave up)

But `createNodePromptExecutor().promptText` returns `{cancelled: true}` only when `typeof value !== 'string'`, so `''` comes back as `{value: ''}`, and a test pins that deliberately: *"treats an empty answer as a value, leaving validation to the caller"*.

Why this is blocking rather than a nit, even though it is only a comment:

- It is on a **public exported type** shipping in a `0.x` minor bump, not an internal note.
- **Three downstream tasks read it.** `ask-policy-interactive-resolver` is the immediate consumer, and its whole job is accepting a pasted transaction hash. An agent reading this doc would reasonably write an "empty means the user aborted" branch that can never fire, and would leave `{value: ''}` unhandled. That is a concrete behavioural consequence, not a stylistic one.
- `AGENTS.md` carries a standing rule about exactly this, because a plausible wrong claim becomes indistinguishable from fact once repeated.

**Fix the DOC, not the behaviour.** I am ratifying the behaviour: `promptText` is a generic text primitive, not a hash-specific one, so treating `''` as a value and leaving validation to the caller is the right call, and the resolver can reject an empty hash with a better message than a generic "cancelled" would give. Just drop the "an empty non-answer" clause, and state positively that an empty string IS a value and callers must validate it. Also record the choice, since Gate 2 correctly flagged it as an unrecorded decision.

## Also fix while you are in there (same class, cheap)

`packages/rocketh/test/prompt-capability.test.ts` has two adjacent JSDoc blocks before the executor test: the one describing the hardhat-deploy / `loadEnvironmentFromStore` path sits immediately above the EXECUTOR test, so it documents a test that is not there, while the real `loadEnvironmentFromStore` test further down has none. Integration tests are documentation in this repo by explicit convention, so move it to the test it describes.

## Ratified, do NOT churn any of these on the retry

- **Decision 4, the hardhat default.** `@rocketh/node` injecting its own prompt when the caller supplied none, so every hardhat-deploy environment reports `canPromptForText()` true by default, is exactly the intent. It is the entire reason this task exists rather than threading through the executor, and ADR 0007 says so. Confirmed wanted.
- **Decision 3, run parameters beat the executor's constructor prompt**, including the consequence that a confirm-only prompt on the run parameters downgrades a text-capable executor prompt to false. Correct, and `injectable-prompt-executor-for-extension-tests` depends on exactly that precedence.
- **Decision 2, a method `canPromptForText()`** rather than a readonly boolean. Fine, and worth keeping now that three later tasks name it.
- **Decisions 1 and 5**, the field name `promptExecutor` and leaving `@rocketh/web` entirely untouched with the web-shaped case covered by a local confirm-only fake. Both right. Not importing web's real object keeps `rocketh`'s tests free of a dependency they should not have.
- The pre-existing confirm bug being **captured as an observation rather than silently fixed** was good judgement: changing confirm's behaviour is out of this task's scope and behaviour-identical for both current call sites.

## Verified independently

- **Purely additive widening**: `promptText?` is optional, `prompt` unchanged, every existing implementation still compiles. `canPromptForText()` becoming a required `Environment` member breaks nothing, because the only other `Environment` literal in the repo was the fabricated stand-in deleted in #76.
- **All three construction paths covered**, and the one that mattered is proven end to end: `loadEnvironmentFromFiles` yields `canPromptForText() === true` by default, and a caller-supplied capability-less prompt flips it to false. That is the hardhat criterion, and it is the whole point of the `ExecutionParams` seam.
- **The `request.name` regression guard is real and would fail** if someone wrote `answer.proceed`: it mocks `prompts` returning `{txHash: '0xdeadbeef'}` for a request named `txHash`.
- **Per-capability, not per-executor** is tested with the load-bearing case: a prompt object that EXISTS but has no text method reports false.
- **Nothing branches on it yet**, pinned by a test asserting `auto` still throws without asking the prompt. That was an explicit criterion and it is honoured.
- **Lockfile**: zero diff lines, zero `my-rocketh-project` occurrences.
- Done-move is `backlog → done` (R100), correct for an `--allow-backlog` drive: the task was never promoted into the pool.

wighawag:
Approved. Both Gate-3 blockers are fixed on disk: TextPromptAnswer's JSDoc now states positively that an empty string is a VALUE the caller must validate (rocketh-core/src/types.ts), matching createNodePromptExecutor().promptText and its pinning test, and it is recorded as decision 6; the stray hardhat/loadEnvironmentFromStore doc block now sits on the loadEnvironmentFromStore test, with the executor test carrying its own. Claim-vs-reality: the widening is additive (optional promptText beside an unchanged confirm), promptExecutor rides ExecutionParams and its resolved form by identity through resolveExecutionParams, and all three createEnvironment call sites are reached and tested (executor, loadEnvironmentFromStore, the shared harness, the last via the existing partial pass-through with no new harness API). Node implements promptText keyed off request.name with the regression test using 'txHash', web is untouched, and the pre-existing confirm .proceed bug is captured as an observation rather than fixed. Cleanup-vs-behaviour: the slice is inert, pinned by a test asserting UnknownSignerError still throws under 'auto' with the prompt never consulted. Coherence: canPromptForText() is per-capability, not per-executor, and 'capability' is used as ADR 0006 already uses it. Destination check: merged as written we get a capability the environment carries on every construction path, hardhat included by default, with nothing branching yet, which is exactly the EXPAND step ADR 0007 and the task describe. Three non-blocking findings, the first worth reading before ask-policy-interactive-resolver starts: the predicate has no TTY probe, so a non-TTY CI run reports true.

wighawag:
APPROVE ✅ — Gate-3 re-review (my BLOCK is resolved)

Both doc-vs-reality fixes landed, and the first one landed better than I asked for.

- **`TextPromptAnswer`** no longer claims an empty answer is a cancellation. It now states positively that an empty string IS a value (`{value: ''}`), that this is a GENERIC text primitive which does not judge the text, and that the caller validates, with the resolver case spelled out: reject `''` and anything malformed, then decide re-ask / abort / defer. That is exactly the guidance `ask-policy-interactive-resolver`'s agent needs, and it now matches the code.
- **Behaviour is unchanged**, which is what I ratified: `promptText` still returns `cancelled` only when the key is absent, so `''` is still a value and its test still pins it.
- **The stray JSDoc** now sits above `reaches the environment through loadEnvironmentFromStore`, the test it actually describes.
- The choice is recorded in `decisions-prompt-capability-on-the-environment-2026-08-10.md`, closing Gate 2's unrecorded-decision finding.

Re-verified on the updated tip: lockfile zero diff, and the ratified decisions were not churned.

Everything from my first pass stands: purely additive widening, all three construction paths threaded with the hardhat one proven end to end (`canPromptForText()` true by default through `loadEnvironmentFromFiles`, false with a capability-less prompt), the `request.name` regression guard that would genuinely fail if someone wrote `answer.proceed`, per-capability rather than per-executor checking tested with the prompt-exists-but-cannot-ask case, and the explicit guard that nothing branches on the capability yet.

Merging. This unblocks `ask-policy-interactive-resolver`.

PR #80, opened 2026-08-10
feat(ask-policy-interactive-resolver): Add the 'ask' policy and the int…

wighawag:
Task: `work/tasks/done/ask-policy-interactive-resolver.md`

Done. All checks green (`pnpm typecheck`, `pnpm build`, `pnpm format:check`, `pnpm test` = 367 passed).

## What this rebuild changed (the three Gate‑3 blocks)

**1. CI can no longer take the ask path — the gate is in the RUNTIME, not the predicate**
`packages/rocketh-node/src/environment/prompt.ts`: `createNodePromptExecutor()` now supplies `promptText` **only when `process.stdin.isTTY`** (probe injectable for tests). `canPromptForText()` is untouched and stays pure method presence (ADR 0007), so the three dependent tasks keep the shape they expect; a non‑TTY run simply has no capability and falls through the already‑tested degrade path. `loadEnvironmentFromFilesWithSpecificConfig` (the hardhat-deploy path) now builds its prompt **per call** rather than reusing the import-time singleton, so the ability reflects the run's stdin.

**2. The non-TTY behaviour of `prompts` is MEASURED, not assumed**
`docs/spikes/ask-policy-interactive-resolver/prompts-non-tty-behaviour.md` + a re-runnable `probe.mjs`. `prompts@2.4.2`: stdin `/dev/null` → the promise **never settles** (node exits silently); open pipe → **hangs**; piped bytes → read as keystrokes. It **never rejects**, so decision 4's "prompt threw, defer instead" could never have rescued a CI run. Only not asking can. Also recorded: the CONFIRM prompt has the same failure mode (out of scope, captured in `work/notes/observations/confirm-prompt-non-tty-2026-08-10.md`).

**3. The pasted-hash wait is bounded, and happens once**
`requireSuccessfulExecutedTransaction` → `waitForPastedTransaction` (`packages/rocketh/src/environment/index.ts`). The bound is on *"this node has never heard of that hash"* (`PASTED_TRANSACTION_LOOKUP_ROUNDS = 10` rounds of `eth_getTransactionByHash` at the run's polling interval), **not** on mining: once the node knows the transaction the wait is the ordinary unbounded one, confirmations included, so a slow Safe execution still resolves. Giving up names the hash as not found and reprints the transaction still to execute; nothing saved, nothing tracked. The receipt it fetches is handed to `savePendingExecution`/`savePendingDeployment` through a one-shot map, so one pasted transaction is waited for once.

**Docs**: the capability paragraph now states the TTY rule and why; a new paragraph says a **DEPLOYMENT** pauses and asks the same way today, that the deployed address is not verified yet, and to prefer the defer workflow until `interactive-deployment-address-recovery` lands. Changeset extended (adds `@rocketh/node: minor`).

Ratified items were left alone: decision 5, decision 3's prompt contract, decisions 2/4/6, the doc-only `@rocketh/unknown-signer` change, the pending-state pipeline, tracker registration, and every signable-class / raw-transaction test.

## Mutation table (each assertion verified to actually fail)

| mutation applied to production code | tests that went red |
| --- | --- |
| always supply `promptText` (drop the TTY gate) | `does NOT supply promptText when stdin is not a TTY`, `defaults to the process stdin…`, `withholds the capability when stdin is not a TTY…` |
| loader reuses the import-time singleton prompt | `gives the environment the capability by default on the hardhat-deploy path` |
| remove the handed-over-receipt short-circuit in `waitForTransaction` | `does not wait for the same pasted transaction twice`, `does not reuse a handed-over receipt…` |
| keep the receipt in the map (drop the one-shot `delete`) | `does not reuse a handed-over receipt for a later transaction` |
| make the lookup unbounded (`round >= Infinity`) | `gives up on a hash this node has never seen…` (5s timeout, i.e. the hang) |
| bound the MINING wait too (poll the receipt in phase 1) | `gives up on a hash…`, `keeps waiting for a known-but-unmined…`, both wait-once tests |

No assertion passes because two branches coincide: the two `auto` directions assert opposite observables (a receipt + a consulted prompt vs an error + an untouched prompt), and the unknown-hash test asserts the *bound* on lookup count as well as the message.

## Decisions recorded

`work/notes/observations/decisions-ask-policy-interactive-resolver-2026-08-10.md` gains **7** (the TTY refusal: what it touches, alternatives rejected, the measurement) and **8** (the not-found refusal + the handed-over receipt), and decision 1's OPEN item is now marked resolved. `work/notes/observations/review-nits-…` gains a dated resolution block for all five nits. Each also carries JSDoc at its choice site.

## Files touched

`packages/rocketh/src/environment/index.ts`, `packages/rocketh/test/interactive-unknown-signer.test.ts`, `packages/rocketh-node/src/environment/prompt.ts`, `packages/rocketh-node/src/executor/index.ts`, `packages/rocketh-node/test/{prompt-executor,prompt-capability-through-loader}.test.ts`, `documentation.md`, `.changeset/ask-policy-interactive-resolver.md`, two observation notes (+1 new), `docs/spikes/ask-policy-interactive-resolver/{prompts-non-tty-behaviour.md,probe.mjs}`.

Working tree contains only these changes; no git operations performed, and the task body was not touched.


wighawag:
Approved. Claim-vs-reality: every premise the task leans on holds on disk. The policy union gains ask in @rocketh/core, resolveUnknownSignerBehaviour is a pure (policy, capability) function with both auto directions and the explicit-ask ceiling pinned by unit tests, and the seam consults canPromptForText() in exactly one place, still inside the unsignable branch. Cleanup-vs-behaviour: all acceptance criteria are met, including the load-bearing absence assertion (no eth_sendTransaction / eth_sendRawTransaction on the resolved path), the tracker registration, the non-success receipt refusal saving nothing, the two-step single-run action, local/node/impersonated/raw untouched, and a throw frame beating an ambient ask. Tests build a real environment on an in-memory store, so no shared location is written. Coherence: ask, capability-as-ceiling and the policy/autoImpersonate split match ADR 0006 and 0007; the new UnknownSignerBehaviour type is a genuinely distinct concept, not a fork of policy. Destination check: spec stories 1, 2, 4 and 5 land and are provably exercised; 3, 7 and 8 are other backlog tasks. The one wrinkle is story 6: the resolver is deliberately not gated to executions, so deployments now resolve interactively BEFORE the address invariants exist, and a deterministic deploy trusts expectedAddress verbatim. That is recorded, reversible and owned by the next task, so it is a ratification finding rather than a block. Findings 1 and 3 deserve a human decision before a release.

wighawag:

Excellent engineering. The mutation table is the right way to answer the tautology trap and I checked its central claim independently: the policy really is consulted only inside the `unsignable` branch, and the three signable classes plus the pre-signed `raw` path each have a test proving they never prompt. The 23 tests cover every criterion including the two-step-in-one-run case that is the whole point of resolving rather than throwing.

I am blocking on ONE thing, and it is an acceptance criterion, not a nit.

## Blocking: this task covers story 5, and for the realistic CI case it does not hold

Criterion: *"With NO text capability, `ask` degrades to `throw`. It never prompts and never hangs, so a CI run cannot block (covers story 5)."* That is tested and passes, but it only covers the case where the capability is absent. The case that matters is the one this PR creates:

`canPromptForText()` is pure method-presence with no TTY probe, and `@rocketh/node` injects `promptText` on BOTH the CLI and loader paths. So a CI job running the `rocketh` CLI, or hardhat-deploy, now has the capability, and the default `auto` policy takes the **ask** path where it previously threw. Whether that job hangs then depends entirely on what the `prompts` library does against a non-TTY stdin, which the PR does not claim to have verified. Decision 4 saves the run only if `prompts` REJECTS; if it blocks on an open stdin, the run hangs.

Two things make this blocking rather than a follow-up:

1. **It is a behavioural regression risk introduced by this diff.** Before it, `auto` always threw in CI. After it, CI behaviour depends on unverified library behaviour against a pipe. Story 5 exists precisely to prevent that, and this task claims to cover it.
2. **The new documentation asserts the opposite.** The section added here tells the reader CI never blocks on a prompt. That sentence is not currently true, and it is the third time in this drive I have had to bounce a doc claim that contradicts the code.

I know Gate 2 raised this as non-blocking and that the capability was task 1's to define. I disagree on placement, not on analysis: task 1 could not have known, because nothing branched on the capability then. This diff is what makes it live, so it is this task's to close.

**The fix I recommend, which preserves ADR 0007.** Do NOT put a TTY probe inside `canPromptForText()`: keeping the predicate pure method-presence is the decision ADR 0007 records, and three later tasks depend on it. Instead, have **`@rocketh/node` supply `promptText` only when stdin is actually a TTY**. Runtime knowledge belongs in the runtime, the capability stays a clean method-presence check, and a non-TTY CI run simply has no text capability, so it falls through the already-tested degrade path. Verify the non-TTY behaviour of `prompts` while you are there and record what you found, so the next person does not have to re-derive it.

## Also fix, cheap and in the same module

**A pasted hash this node has never seen hangs the run forever.** `requireSuccessfulExecutedTransaction` calls `waitForTransaction`, whose receipt polling has no deadline, and then `savePendingExecution` waits on the SAME hash a second time. For a normal broadcast that is fine because the transaction was definitely sent; for a hash a human typed it may never exist (wrong chain, a plausible typo), and the user gets an unbounded spinner with Ctrl-C as the only exit, then a second wait for the same transaction. Bound the wait for the pasted-hash case with a message naming the hash and saying it was not found, and stop waiting twice for one transaction.

## Answers to the questions the PR and Gate 2 asked

- **Decision 5, the resolver not being gated to executions: RATIFIED, ship the intermediate state.** A deployment pausing and asking is a strict improvement over throwing, nothing releases between this and `interactive-deployment-address-recovery`, and artificially gating it would only force the next task to un-gate. I am planting a note on that task so its agent does not read "the blocking task solved the deployment case" and bounce; that is my job, not yours.
- **Nit 2, the doc not mentioning deployments: yes, add the sentence.** One line saying a deployment also pauses and asks today, and that address verification arrives with the follow-up. An accurate doc about a temporary state beats a doc that silently omits it.
- **Decision 1's open question is answered by the fix above**, so drop it from the unresolved list once the TTY gating lands.
- **Decision 3's prompt contract RATIFIED**: empty means defer, "cannot sign" matched case and separator insensitively, malformed re-asked at most 3 times. Sensible defaults, and `per-call-ask-override-and-deferral-precedence` inherits them.
- **Decisions 2, 4 and 6 RATIFIED** as written.

## Verified independently, do not churn

- Policy consulted only inside the `unsignable` branch, with `local`, `node`, `impersonated` and pre-signed `raw` each pinned by a test asserting the prompt is never consulted. This is the invariant that drifted through three documents and bounced a task set, and it holds here.
- The pasted hash flows through the UNCHANGED pipeline: nothing about pending state or receipt waiting was reimplemented, and a test asserts no send RPC was attempted.
- Tracker registration is present and mutation-tested.
- The `@rocketh/unknown-signer` change is doc-only and correctly does not encroach on `per-call-ask-override-and-deferral-precedence`.
- Lockfile: zero diff, zero `my-rocketh-project`.

wighawag:
Approved. Claim check: every acceptance criterion is delivered by the bytes on disk. The policy union gains ask, resolveUnknownSignerBehaviour is a pure function with the full truth table unit-tested in both directions, and the seam consults canPromptForText() in exactly one place, still inside the unsignable branch, with local, node, impersonated, raw and mixed-run cases pinned so ADR 0006 cannot drift again. The pasted hash flows through the untouched savePendingExecution pipeline, the absence of eth_sendTransaction / eth_sendRawTransaction is asserted, the two-step action completes in one run, and non-success or absent status fails before anything is saved or tracked. Cleanup vs behaviour: the @rocketh/unknown-signer change really is doc-only, and the frame still forces throw over ask and never over impersonation. All three requeue fixes landed and are honest: the TTY gate sits in the runtime, not in canPromptForText(), which stays pure method presence per ADR 0007; the prompts non-TTY behaviour was measured with a reproducible probe rather than assumed; the unknown-hash lookup is bounded while mining stays unbounded; and the fetched receipt is handed over once via a consumed map. Coherence: capability-as-ceiling, policy vs node capability, and the new UnknownSignerBehaviour type all fit the existing language. Destination check: merged as written, an unsignable from at a keyboard pauses, resolves and continues in one run, and a run with no terminal behaves exactly as before. The deployment funnel now asks too, which is ratified and documented but untested here; the follow-up task owns it. Findings are ratification and hygiene only.

wighawag:
APPROVE ✅ — Gate-3 re-review (my BLOCK is resolved, and it was justified)

All three fixes landed, and the investigation I asked for proved the block was not theoretical.

## The blocking issue, closed properly

`@rocketh/node` now attaches `promptText` **only when `process.stdin.isTTY === true`**, so the method is genuinely ABSENT in a non-TTY run and `canPromptForText()` falls through the already-tested degrade path. Critically, `canPromptForText()` stayed pure method-presence, so ADR 0007 is intact and the three later tasks that branch on it are undisturbed. The injectable `isStdinInteractive` seam is the right way to test both sides.

And the measurement answers the open question: `prompts` asked a question with no terminal behind it **never answers and never fails**. So the pre-fix code would have HUNG CI, and decision 4's throw-degrades-to-defer would NOT have saved it. That is exactly the failure story 5 exists to prevent, and it is now recorded rather than assumed.

The other two fixes are good: the unknown-hash lookup is bounded and fails with a message naming the hash as not found on this network, the receipt is cached so the pipeline no longer waits twice for one transaction, and the docs now say a DEPLOYMENT also pauses and asks today with address verification still to come, advising the defer workflow for deployments meanwhile. The corrected CI paragraph explains the mechanism rather than just asserting safety.

## Correcting Gate 2 on `docs/spikes/`

Nit 3 asks whether `docs/spikes/` is "a brand-new convention to pin in `CONTEXT.md`". It is not new: `work/protocol/WORK-CONTRACT.md` conflict-safety rule 8 already mandates exactly this, and even names the artifact types. Tasks and specs must NOT carry a co-located `<slug>/` sidecar because they FLOW between status folders and the sidecar strands; their durable companion artifacts belong in "a STABLE, NON-FLOWING home", and the contract says "the convention is `docs/spikes/<slug>/`", listing "a build/measurement script" and "a captured measurement" among what goes there. A probe script plus its captured measurement is the textbook case. **The author followed the contract correctly; no change needed.**

The half of that nit worth keeping: the measured `prompts@2.4.2` behaviour is verified EXTERNAL ground truth, which `CONTEXT.md` homes in `work/notes/findings/` with a `source:`. A short finding pointing at the spike artifact would make it discoverable as domain knowledge rather than as one task's evidence. Worth doing, not worth blocking.

## Surfacing two residuals rather than deciding them

- **A pty-allocating CI still gets the capability.** `docker run -t` and some Jenkins/GitLab configurations allocate a terminal, so `isTTY` is true and an unsignable `from` under `auto` would pause for input nobody types. The docs are honestly qualified ("a CI job, whose stdin is not a terminal"), so this is a residual and not a false claim. Whether to also withhold the ability when `process.env.CI` is set is a real judgement call, and I am not making it unilaterally: TTY-only is defensible (someone running `docker run -t` locally IS interactive), and the `CI` variable is a heuristic. Flagging for you.
- **The 10-round bound is tight.** Ten rounds at the default 1s polling interval is about ten seconds before a hard failure. A hash pasted straight after a Safe execution against a load-balanced public RPC can plausibly take longer to become visible, and it bites hardest on the deployment funnel, where giving up leaves no record. Recovery exists (re-run and paste the same hash), but a wall-clock grace period, or simply a larger bound, would fit the real-world scenario better. Your call on the number.

Neither blocks: nothing releases between now and your decision, and both are one-constant changes.

## Verified independently, first pass and this one

Policy consulted only inside the `unsignable` branch, with `local`, `node`, `impersonated` and pre-signed `raw` each pinned by a test asserting the prompt is never consulted. The pasted hash flows through the UNCHANGED pending-state pipeline with no send RPC attempted. Tracker registration present and mutation-tested. The rejected alternative for decision 5 is recorded at the choice site, which is exactly the standard the repo asks for. Lockfile: zero diff, zero `my-rocketh-project`.

The mutation table remains the right discipline; please keep extending it.

One stale reference to tidy whenever convenient: the decisions note still names `requireSuccessfulExecutedTransaction` in item 2, but the function that landed is `waitForPastedTransaction` (item 8 uses the new name). Not blocking, since observations are append-only and an `## Update` fixes it.

Merging. This unblocks `interactive-deployment-address-recovery` and `injectable-prompt-executor-for-extension-tests`.

PR #81, opened 2026-08-10
feat(injectable-prompt-executor-for-extension-tests): Injectable fake p…

wighawag:
Task: `work/tasks/done/injectable-prompt-executor-for-extension-tests.md`

Done. All checks green.

## What I did

**Reality check first (no drift):** `ExecutionParams.promptExecutor` / `ResolvedExecutionParams.promptExecutor` carry the capability verbatim (`packages/rocketh-core/src/types.ts:365,671`, passed through at `packages/rocketh/src/executor/index.ts:306`), `env.canPromptForText()` is pure `promptText` method presence (`packages/rocketh/src/environment/index.ts:534`), and `createTestEnvironment` already accepts `executionParams?: Partial<Omit<ExecutionParams,'provider'>>`. So the task's central claim holds: no new harness API was needed.

**New (test-first):**
- `packages/rocketh-unknown-signer/test/interactive-prompt.integration.test.ts` — written and run red before the production code. Three documentation-style tests through the real harness: a canned hash continues the run (real receipt for the pasted hash, zero send RPCs), `'cannot sign'` defers with `UnknownSignerError`, and a capability-absent fake degrades to throw with `requests` empty (never consulted).
- `packages/rocketh-test-utils/src/mock-prompt.ts` — `createMockPromptExecutor`. Scripted `textAnswers` (string → `{value}`, `TextPromptAnswer` passthrough, `Error` thrown), `requests` / `textRequests` / `exited` recording, loud failure past the end of the script, script copied so a shared array is not consumed. Prompt double only: no environment, no broadcast path, and the regrowth-fence test (`^create.*Environment$` → exactly `createTestEnvironment`) still passes.
- `packages/rocketh-test-utils/test/mockPromptExecutor.test.ts` — the double's own contract.
- `.changeset/injectable-prompt-executor-for-extension-tests.md` — `@rocketh/test-utils: minor` (real API addition on a published package). `@rocketh/unknown-signer` is test-only, so it carries no bump (precedent: `.changeset/migrate-deploy-and-read-tests.md`).
- `packages/rocketh-test-utils/src/index.ts` re-export, plus one paragraph in `TESTING.md` so the shared-helpers section does not drift.

## Decisions (recorded durably in the module JSDoc of `packages/rocketh-test-utils/src/mock-prompt.ts`)

- **Capability-absent shape = `createMockPromptExecutor()` with no `textAnswers`** (returns an executor with no `promptText` at all). It mirrors the domain rule under test: absence of the method IS the signal. Alternatives considered and rejected: a second builder (`createConfirmOnlyPromptExecutor`, as `packages/rocketh`'s local tests have) doubles a published package's surface for one line of behaviour; a `canPromptForText` option re-uses the environment predicate's name as a constructor knob. A capability-present-but-answerless prompt is `{textAnswers: []}`. This touches any sibling task that injects the fake, hence the record.
- **Naming `createMockPromptExecutor`**, matching the existing `createMock*` doubles (`createMockProvider`, `createMockArtifact`). `CONTEXT.md`'s "mock environment" caution is about *environments* specifically; this name cannot match the fence regex and does not re-mean that entry.
- **The double stays out of the resolver's vocabulary**: `'cannot sign'` is just a scripted string, not a re-encoded constant (`CANNOT_SIGN_ANSWER` is internal to `packages/rocketh` and not exported), so there is no second source of truth to drift.

## Verification

`pnpm build`, `pnpm typecheck`, `pnpm format:check`, `pnpm test` (31 files / 377 tests) all pass. `pnpm changeset status --since=main` currently reports "no changesets found" only because my changeset file is still untracked (changesets reads it via `git diff` against the merge-base); it becomes visible once the runner commits with `git add -A`. `pnpm test:getting-started` was not run (network-dependent, untouched by this change).

Working tree contains only the five intended files.


wighawag:
Approved. Claim check: every premise the task rests on still holds on disk. promptExecutor rides ExecutionParams (rocketh-core/src/types.ts:365), promptText is the optional method whose absence is the capability signal (types.ts:857-869, ADR 0007), and createTestEnvironment's partial pass-through (test-environment.ts:87) took the fake with NO new harness option: src/index.ts gains only the re-export, test-environment.ts is untouched. Cleanup/conformance: all acceptance criteria land. Canned hash, cannot-sign, and the capability-absent shape are all expressible; requests / textRequests make both the positive and the negative assertion (nobody was asked) read plainly, which the sibling catchUnknownSigner task needs. No environment is fabricated and no broadcast path reimplemented; the regrowth fence still sees exactly one builder, since the new export ends in Executor, not Environment. Changeset carries a minor bump on the published package with a clear purpose line. Coherence: terminology matches the CONTEXT.md test-environment vs mock-environment entry and the createMockProvider / createMockArtifact family; confirm auto-proceed mirrors the web prompt rather than re-meaning it. Destination check: merged as written, extension packages can drive the interactive unknown-signer path with no TTY through the existing route, with a working documentation example. Findings are ratification items only, chiefly the omit-vs-empty wording in the published changeset and the unscriptable confirm half.

wighawag:

## Parallel-build safety verified, not assumed

This was built in parallel with #82 off a `main` that did not yet contain it, so its green predated that merge. I rebased it onto the merged `main` in a **throwaway clone** (your checkout untouched) and ran the full gate on the combined tree: clean rebase, **32 test files / 385 tests passing**. The two diffs are genuinely disjoint, which is what made the parallel run safe, and it is now confirmed rather than argued.

## Criteria

- Canned-hash and "cannot sign" answers, plus the capability-absent shape, plus recorded requests so a test can assert **that nobody was asked** — which `per-call-ask-override-and-deferral-precedence` needs to prove `catchUnknownSigner` defers without consulting a prompt. All present.
- **The canary held: `createTestEnvironment` is untouched.** No harness option was added; the fake rides the existing `executionParams.promptExecutor` pass-through. That was the criterion I set specifically to detect whether the ADR 0007 plumbing had regressed, and it says the capability route works as designed.
- No environment is fabricated and no second builder appears, so the regrowth fence from #76 stays satisfied.
- `minor` bump on a published package for a new export, correctly.
- Lockfile: zero diff.

## Ratified

The capability-absent default, the deliberately non-scriptable confirm half (mirroring `@rocketh/web`), an exhausted script throwing (which the resolver catches and defers, so a mis-scripted test surfaces as an `UnknownSignerError`), and the extra `Error` answer shape for the no-terminal case. All defensible, all now documented in the nits file.

## Follow-ups, none blocking

- **Consumer-facing wording, omit vs empty.** `createMockPromptExecutor()` with options OMITTED is capability-absent, while `textAnswers: []` is capability-PRESENT but exhausted. The module doc is precise; the changeset (which becomes the published CHANGELOG) and `TESTING.md` both say "with no scripted answers", which a reader will reasonably apply to the empty array and get the opposite behaviour. I am folding this into the release-prep pass that already has to tighten two other changeset clauses before `changeset version` runs, rather than spending a build cycle on wording.
- **No `## Decisions` block was recorded at build time**, so the four decisions above had to be reconstructed from source by the review. The task template asks for it; worth honouring next run.
- The end-to-end example sits in `packages/rocketh-unknown-signer/test/` but drives `env.broadcastExecution` directly and never calls `catchUnknownSigner`. It satisfies the criterion (an extension-package test proving the shared route), but its home implies coverage of that package's own API which it does not add.
- **Vocabulary divergence worth pinning:** the shared double is `createMockPromptExecutor`, while `packages/rocketh/test` calls the same two notions `createScriptedPrompt` and `createConfirmOnlyPromptExecutor`. The duplication is intended (the dependency edge), the naming split is not, and this repo has already paid for unpinned vocabulary once. A `CONTEXT.md` line would be cheap.

Merging.

PR #82, opened 2026-08-10
feat(interactive-deployment-address-recovery): Recover and verify the d…

wighawag:
Task: `work/tasks/done/interactive-deployment-address-recovery.md`

All checks green. Here's what I did.

Confirmed against current reality, no drift:
- The resolver landed at the shared choke point and is **not** gated to executions (`packages/rocketh/src/environment/index.ts`, the `unsignable` branch of `broadcastTransaction`) — expected per the forward-pointer.
- The shared receipt-**status** check is present (`waitForPastedTransaction`), and the address verification had **not** landed.
- The line the task told me to read before designing: `packages/rocketh/src/environment/index.ts:892` (now :919) in `waitForDeploymentTransactionAndSave` — `const contractAddress = pendingDeployment.expectedAddress || receipt.contractAddress;` followed by `if (!contractAddress)`. That is exactly the hole: the zero address is truthy, so it sails through into a saved record.

## What changed

`packages/rocketh/src/environment/index.ts`
- `broadcastTransaction`'s `origin` parameter is now **required** and a **discriminated union** (`BroadcastOrigin`: `{type:'execution', contract?}` | `{type:'deployment', name, expectedAddress?}`), so the check cannot be skipped by a future caller. Still a private closure, no exported surface.
- New `requireDeployedContract`, run in the interactive branch immediately after the status check and **before** the pending-transaction write, the gas-tracker push and `save`: expected address present → require **code at it** (`eth_getCode`, never transaction parsing, receipt's own address ignored); otherwise → receipt `contractAddress` must be present **and not the zero address**. Every failure names the deployment, the pasted hash and the transaction that still needs executing.

`packages/rocketh/test/interactive-deployment-address.test.ts` (new, written before the production code; real environment built locally, mock provider, drives `env.broadcastDeployment`, no `@rocketh/deploy` import): 8 tests — receipt-address recovery + save under name with no send; code-at-expected-address confirmation (receipt deliberately reports a *different* address); no-code failure; absent address; zero address; reverted receipt on the deployment path; and two anti-regression tests pinning that a signable-account deploy (deterministic included) triggers **no** `eth_getCode` and no prompt.

`documentation.md` — replaced the "verification is coming" paragraph in the existing "Handling unknown signers" section with what the deployment path now guarantees.

`.changeset/interactive-deployment-address-recovery.md` — `rocketh: patch`. (`changeset status --since=main` is red only because it reads changed files via `git diff`, which cannot see an untracked file; it will be green once the runner commits.)

## Tests can actually fail (mutation-checked)

Before the production code: 4 of 8 failed (the reverted-receipt one passed, as expected — it inherits the shared status check). After, I mutated and reverted three times: dropping the zero-address clause → only the zero test fails; checking code at the receipt's address instead of the expected one → only the deterministic test fails; running the check on the normal path too → the two anti-regression tests plus two others fail.

## Decisions (recorded durably)

`work/notes/observations/decisions-interactive-deployment-address-recovery-2026-08-10.md`, six entries, each also carrying a JSDoc at its choice site: (1) required discriminated `origin`; (2) the new refusal and where it runs; (3) expected address wins and the receipt's is ignored; (4) an unanswerable `eth_getCode` fails rather than being assumed fine; (5) normal deploys deliberately **not** tightened; (6) a coherence note that `origin` already means "sender address" on `PendingTransaction.transaction.origin` — pre-existing, flagged not introduced.

## Verification

`pnpm typecheck`, `pnpm build`, `pnpm format:check` all pass; `pnpm test` 375 passed / 30 files, including the execution-path tests from the blocking task untouched. Working tree contains only the four intended files plus the decisions note.


wighawag:
Approved. Claim-vs-reality: the diff does what the task specified, and the premises check out on disk. The address invariants run at the shared choke point immediately after the status check and BEFORE the pending-transaction write, the gas-tracker push and any save, so a bad paste leaves no state; the receipt that was validated is the same one reused for the save via the pasted-receipt cache, so the checked address is the recorded address. The deterministic branch is reachable in production: @rocketh/deploy passes expectedAddress for create2/create3 (packages/rocketh-deploy/src/index.ts:438,478). Cleanup-vs-behaviour: every acceptance criterion is met, including both no-usable-address shapes (absent and the truthy zero address, lowercase-normalised), the reverted-receipt case asserted on the deployment path rather than inherited, and an anti-regression test asserting the ABSENCE of eth_getCode for a signable deterministic deploy. Execution-path files are untouched in the commit. Docs extend the existing unknown-signer section and drop the stale 'verification is coming' paragraph; a changeset is present. Composition: the widened parameter is a private closure with two in-module callers, so no public surface moves. Destination check: story 6 of unknown-signer-interactive is delivered, an interactively-deferred deployment now records an address it either read from the pasted receipt or confirmed on-chain, or it fails naming the deployment, the hash and the transaction still owed. Four non-blocking ratification/coherence findings, chiefly the untested eth_getCode-error refusal and the two meanings of 'origin'.

wighawag:

Every criterion met, and the two that mattered most are met with DISCRIMINATING tests rather than merely present ones.

- **Verification runs before anything is recorded.** `requireDeployedContract` sits after the status check and before the tracker push, the receipt cache and the return into the save pipeline, so a failed verification genuinely leaves no state behind. I checked the ordering in the source rather than trusting the description.
- **Code-at-expected-address is proven, not assumed.** The deterministic test deliberately has the receipt report a DIFFERENT address, so it can only pass if the expected address plus the code check is what decided. That is the test I would have asked for.
- **All three failure shapes covered separately**: no code at the expected address, absent contract address, and the **zero address**, which is the one that mattered because the previous `expectedAddress || receipt.contractAddress` then `if (!contractAddress)` guard let a truthy zero address sail into a saved record. Plus a reverted receipt pinned on the deployment path specifically, so it cannot bypass the shared invariant.
- **No new failure mode for normal deploys**, pinned by two anti-regression tests asserting a signable-account deploy (deterministic included) triggers no `eth_getCode` and no prompt. That was an explicit criterion and it is honoured with real assertions.
- Tests are in `packages/rocketh/test/`, drive `env.broadcastDeployment`, and import no `@rocketh/deploy`, so the nx cycle stays open. Lockfile: zero diff, zero `my-rocketh-project`.

**The forward-pointer did its job.** The drift check correctly read "deployments already resolve interactively" as expected rather than as drift, and built on the existing path instead of bouncing or re-gating. Recording that because it validates planting the note.

## Ratifications

- **`origin` becoming a REQUIRED discriminated union** on the private choke point. This is the strongest form of my "the check cannot be skipped by a future caller" criterion, and Gate 2 confirmed by grep that `broadcastTransaction` is module-private with exactly the two callers, so there is no public surface change. Ratified.
- **Ignoring the receipt's own `contractAddress` when an expected address exists**, even when present. The rationale (a factory receipt names the factory call, and the expected address is the recorded one) is sound and pinned by the different-address test. Ratified, and the deliberate asymmetry is worth the comment it already carries.
- **Failing on an unanswerable `eth_getCode`** rather than treating an RPC error as no-code or as fine. Right direction: fail loud. Ratified.

## Two small gaps, neither blocking

- That last refusal is **the only new one with no test**, and `documentation.md` lists the other three failure shapes but not it. Worth a test and one clause whenever this area is next touched.
- Gate 2's coherence point is a good catch: `origin` now carries two meanings in one module, the choke point's what-produced-this bag and `PendingTransaction.transaction.origin`, which is a SENDER address. Neither is pinned in `CONTEXT.md`. A glossary entry before a third meaning appears would be cheap insurance, and this repo has already paid for unpinned vocabulary once.

Merging. This unblocks `per-call-ask-override-and-deferral-precedence`.

PR #83, opened 2026-08-10
feat(per-call-ask-override-and-deferral-precedence): Per-call policy ov…

wighawag:
Task: `work/tasks/done/per-call-ask-override-and-deferral-precedence.md`

Done. Everything is green and the working tree is clean.

Verified against the code, not the task's snapshot:

- **Policy-frame stack** exists with the assumed shape: `createUnknownSignerPolicyStack` (`packages/rocketh/src/environment/unknownSignerPolicy.ts`), exposed as `pushUnknownSignerPolicy(frame)` / `popUnknownSignerPolicy()` on the core `Environment` interface (`packages/rocketh-core/src/types.ts`), frame is an OBJECT (`UnknownSignerPolicyFrame = {policy}`).
- **Interactive value + capability predicate landed**: `UnknownSignerPolicy = 'throw' | 'ask' | 'auto'`, `resolveUnknownSignerBehaviour(policy, {canPromptForText})`, `env.canPromptForText()`, resolver at `interactiveUnknownSigner.ts`.
- **Per-call override did NOT already exist**: the only `pushUnknownSignerPolicy` caller in any `src/` was `catchUnknownSigner`. No drift, so I built it.
- `catchUnknownSigner` still takes a THUNK (compile error + runtime guard) — not loosened.

## What changed

- `packages/rocketh-unknown-signer/src/index.ts` — new export **`withUnknownSignerPolicy(env)(policy, action)`**: pushes a policy frame for one action, returns the action's value, propagates its throw. Both wrappers now share ONE push/pop site (`runUnderPolicyFrame`, pop in `finally`), and the thunk guard names the wrapper you actually called. `catchUnknownSigner`'s behaviour is unchanged.
- `packages/rocketh/src/environment/unknownSignerPolicy.ts` — module JSDoc: both wrappers, one precedence rule, capability still the ceiling over a frame.
- `packages/rocketh/test/unknown-signer-policy-precedence.test.ts` (new, real env + local mock provider) — frame beats run-level, run-level beats chain-level (both directions), chain-level beats the default, `ask` frame degrades with no text capability, no stranded frame on the throw path, and `local`/`node`/`impersonated`/`raw` unaffected with an `ask` frame in force.
- `packages/rocketh-unknown-signer/test/per-call-policy.integration.test.ts` (new) — per-call override both directions, capability ceiling (two shapes), frame balance on throw, LIFO nesting, thunk guard, the fork rehearsal end to end (story 3), and the **deferral guarantee**: under an ambient `'ask'` with a working prompt, `catchUnknownSigner` defers with `promptExecutor.requests === []` and a deferral message byte-identical to the ambient-`'throw'` one.
- `documentation.md`, package `README.md`, `.changeset/per-call-ask-override-and-deferral-precedence.md` (`@rocketh/unknown-signer`: minor).

## Tests are discriminating (verified by mutation, mutants reverted)

- `effective()` ignoring frames → 2 precedence tests red.
- `withUnknownSignerPolicy` not pushing → 5 red; `catchUnknownSigner` not pushing → the 3 deferral-guarantee tests red.
- Capability ceiling removed from `resolveUnknownSignerBehaviour` → my ceiling tests red **only after** I strengthened them to assert the run never showed the `... is PAUSED` presentation ("it threw `UnknownSignerError`" cannot tell a ceiling from a crash, since a broken prompt degrades to the same error).

## Decisions (recorded durably)

`work/notes/observations/decisions-per-call-ask-override-and-deferral-precedence-2026-08-11.md` + JSDoc at the choice site: (1) the override is a **wrapper in `@rocketh/unknown-signer`**, not an option on `deploy`/`execute`/`tx` and not a new `Environment` method; (2) it accepts the whole `UnknownSignerPolicy` union incl. `'auto'`; (3) user-visible: LIFO means an explicit override written *inside* `catchUnknownSigner` wins (the guarantee is about the AMBIENT policy); (4) shared thunk guard/push-pop site; (5) the test technique above.

## Out-of-scope signals captured (not fixed)

- `work/notes/observations/capability-ceiling-tests-survive-a-removed-ceiling-2026-08-11.md` — two sibling tests from `ask-policy-interactive-resolver` survive a removed ceiling.
- `work/notes/observations/executor-comment-says-no-interactive-resolver-exists-2026-08-11.md` — stale comment in `resolveExecutionParams`.

## Gate status

`pnpm format:check` ✅, `pnpm typecheck` ✅, `pnpm build` ✅, `pnpm test` ✅ (34 files, 408 tests). I deliberately did **not** run `pnpm test:getting-started` locally: it re-pollutes the root `pnpm-lock.yaml` (documented observation) and would dirty the tree; it runs in CI. `changeset status --since=main` cannot pass locally because the changeset is still untracked (it diffs against `main`); the check is global (`changedPackages > 0 && changesets === 0`), so the one changeset satisfies it once committed.


wighawag:
Approved. Claim-vs-reality: the diff does what it claims. withUnknownSignerPolicy pushes the EXISTING policy frame (no new channel), the ceiling still runs through resolveUnknownSignerBehaviour at the single seam, and the rocketh-side change is JSDoc only, matching the changeset that bumps only @rocketh/unknown-signer. Cleanup-vs-behaviour: catchUnknownSigner's push/pop moved into a shared runUnderPolicyFrame with the pop in a finally; ordering and balance are still pinned by the pre-existing plumbing tests, so it is a genuine no-op refactor. Acceptance: every criterion has a discriminating test - a frame beating a run policy, a run policy beating a chain policy, both directions, the ceiling asserted by the ABSENCE of the PAUSED presentation (not just the error), frame balance after a throw, signable/impersonated/node/raw paths unchanged under an ask frame, and the deferral guarantee proven by an unconsulted prompt plus a byte-for-byte message comparison. Coherence: one precedence rule, one frame stack, no forked vocabulary; capability stays a ceiling at the seam, not a per-call default. Destination check: spec stories 3 and 8 (including the absorbed core story 9) are delivered end to end, including the fork rehearsal. Findings are ratification items plus one stale invariant comment in the file this commit touched, none of which change behaviour.

wighawag:

The criterion this task existed for is met exactly, and the build did something better than I asked: it mutation-tested the PREVIOUS task's tests and found them wanting.

## The guarantee is finally assertable, and asserted

`defers under an ambient 'ask' WITHOUT consulting the prompt` supplies a prompt that WOULD have answered (`textAnswers: [PASTED_HASH]`) and then asserts `promptExecutor.requests` is empty, plus no `eth_sendTransaction`. That is the discriminating assertion the core slice could not write, and it replaces the tautology the reviewer flagged back then. The companion test compares the deferral message byte for byte between a `throw` run and an `ask` run, which is stronger than checking it merely mentions the transaction, and it protects the invariant that the unwrapped message IS the deliverable.

## The best thing in this PR is the honesty

`capability-ceiling-tests-survive-a-removed-ceiling-2026-08-11` reports that two of `ask-policy-interactive-resolver`'s own ceiling tests **still pass with the capability check removed**, because a run that enters the interactive path with no usable prompt also throws `UnknownSignerError` (the resolver degrades a prompt it cannot call to the defer path). The user-visible difference is that the broken run first shows the human the PAUSED presentation, so the new tests here assert the ABSENCE of that message and genuinely kill the mutant.

That is a real defect in tests I approved two PRs ago. I merged them believing the mutation table; the table was honest about what it killed, but these two assertions were weaker than they looked. Finding it, proving it by mutation, declining to fix another task's tests unilaterally, and capturing it with `needsAnswers` is exactly right. Technique noted as decision 5 and worth adopting as the house style for capability tests.

## Ratifications

- **Decision 1, the per-call override as a WRAPPER** (`withUnknownSignerPolicy(env)(policy, action)`) rather than an `onUnknownSigner` field on `deploy`/`execute`/`tx`. Ratified. The story says "via a call option" and this is a call-SITE option, so it is a mild reading of the wording, but the reasoning is strong: an options field would put the same decision in four packages, each threading it to one choke point, and "which of the four wins" would become a second precedence rule beside the frame stack. My task text pinned "the existing scoped policy-frame stack rather than a new mechanism", which a wrapper is. Flagging it as the one user-visible API shape here that you may want to eyeball.
- **Decision 3, frames nest LIFO so an explicit inner override beats `catchUnknownSigner`'s frame.** Ratified, and the distinction is precise rather than a weakening: the guarantee was always about the AMBIENT policy, and an override the same author deliberately wrote one line down is not ambient. Making the wrapper's frame absolute would fork one precedence rule into "some frames are stronger", which is worse. Documented and pinned by a test.
- **Decisions 2 and 4** ratified: accepting the whole union (so `'auto'` can opt one call back into the capability-aware default) avoids a second vocabulary for the same concept, and sharing one push/pop site plus one thunk guard stops the two wrappers drifting into different scoping rules while keeping the `finally` that balances the throw path in exactly one place.

## Verified

Frame balance on the throw path is tested. The capability ceiling holds for an overridden `'ask'` both with a confirm-only prompt and with no prompt at all, each asserting no PAUSED message rather than just an error. The stale-comment find (`executor-comment-says-no-interactive-resolver-exists-2026-08-11`) is correct and correctly left unfixed as out of scope. Lockfile: zero diff.

Merging. `impersonation-unsupported-hint-and-web-guidance` is the last task in the chain.

PR #84, opened 2026-08-11
feat(impersonation-unsupported-hint-and-web-guidance): Hint when impers…

wighawag:
Task: `work/tasks/done/impersonation-unsupported-hint-and-web-guidance.md`

Done. This was a continuation: I kept the prior attempt's work intact and fixed exactly the three requeue items.

## What I changed (on top of the prior attempt)

**1. BLOCKING fix: the note is now readable (`packages/rocketh-core/src/errors.ts`)**
The impersonation note moved from after `data:` to directly under the `Execute the following transaction out-of-band` header, above `contract:`/`from:`/`to:`/`value:`/`data:`. For an execution `data` is short calldata, but for a *deployment* it is the whole creation bytecode, so the note previously landed thousands of hex characters below where anyone stops reading, in precisely the case the spec targets. The absent-field-means-silence rule is untouched: with `autoImpersonate` off the message is byte-for-byte what it always was (still pinned by the existing test).

Test-first: added `puts the impersonation note ABOVE the transaction fields, where it is still read` in `packages/rocketh-core/test/errors.test.ts`, which builds an error with a 4000-byte `data` and asserts the note's line index (2) sits above every field line. Confirmed RED (`expected 7 to be 1`, then corrected to the true header offset of 2) before the production edit, GREEN after.

**2. Accuracy fix (text-only): candidacy wording**
`needsImpersonationForRun` (`packages/rocketh/src/environment/index.ts:471-486`) requires all three of: named account, absent from `eth_accounts`, and resolved signer is `remote` (the `Signer` union has exactly three variants, `signerOnly | remote | wallet`, `packages/rocketh-core/src/types.ts:575-578`). The old wording, "only NAMED accounts absent from `eth_accounts`", stated a broader rule than the code's. Corrected in the `not-a-candidate` message, its JSDoc, `documentation.md` constraint 1, and the paraphrase line below the constraints, and the changeset text.

**3. Decisions record (was missing)**
`work/notes/observations/decisions-impersonation-unsupported-hint-and-web-guidance-2026-08-11.md`, four entries: the typed `autoImpersonation` union on the published `UnknownSignerErrorData` (ratified, with the accepted cost that a third variant widens a published union), inferring `'attempted'` from the candidate set at the call site (ratified, with the staleness condition: an early-return or filter inside `impersonateAccounts` would make the claim false, and the fix would be to have the helper return `{attempted, succeeded}`), the note-position decision, and the wording correction.

## Not changed, deliberately

`impersonateAccounts`'s swallow is intact, the attempt still happens at setup, the policy is still consulted only inside the `unsignable` branch, and the outcome is read only where the error is built. ADR 0006's capability-vs-policy boundary is unmoved; the anti-regression classification tests in `packages/rocketh/test/impersonation-hint.test.ts` still pin it.

## Checks

`pnpm build`, `pnpm typecheck`, `pnpm format:check`, `pnpm test` (35 files, 420 tests) all pass. Working tree carries only the four modified files plus the new decisions note; the lockfile is unmodified.


wighawag:
Approved. Claim-vs-reality: every documented constraint checks out against the bytes. Candidacy really is named-accounts-only (unknownAccounts is built from Object.values(namedAccounts) filtered by needsImpersonationForRun), autoImpersonate really is run-level (execution param then chain config in executor/index.ts), and @rocketh/web really ships a PromptExecutor with no promptText, so canPromptForText is false and ask degrades to throw. Cleanup-vs-behaviour: the swallow in impersonateAccounts is untouched, the note is computed at setup and read only where the error is built inside the unsignable branch, and the single construction site means the deployment funnel and the interactive prompt get the same message. Coherence: capability-before-seam vs policy-after-seam (ADR 0006) is preserved; no impersonate policy value appears, and the new terms do not re-mean signability or the existing impersonated state. Tests re-pin local/node/impersonated/unsignable and assert the off path is byte-for-byte identical to the pre-change message, which is the strongest guard here. Destination check: merged as written, a user who enabled impersonation against the wrong node is told which of the two silences they hit, the common path gains no noise, and the browser/fork route is written down with its three real constraints. Findings are ratification and polish only: the new exported autoImpersonation field on UnknownSignerErrorData is a public API choice the task did not specify, attempted is inferred from candidacy rather than reported by the helper, and no Decisions record accompanies either.

wighawag:

The mechanism is right, the tests are the best in this chain, and the change is provably control-flow-neutral (zero deletions in `environment/index.ts`, purely additive). I am blocking on placement, which sounds cosmetic and is not.

## Blocking: the hint lands where nobody will read it

In `buildMessage`, the note is pushed AFTER `data:`. For an EXECUTION that is short calldata and fine. For a **DEPLOYMENT** `data` is the full creation bytecode, so the hint sits thousands of characters below the point any human stops reading.

That defeats this task's only deliverable, in precisely the case the whole spec targets. The repo's standing invariant is that the unwrapped throw is the PRIMARY deferral workflow and the error MESSAGE is the deliverable, not a summary. A message that technically says "auto-impersonation was attempted and this node does not implement it", 50,000 characters after the reader has given up, does not functionally say it. And you added a test asserting the note carries through the deployment funnel, so this is exactly the path you covered.

**Fix:** move the note UP, above `from:`/`to:`/`value:`/`data:`. Directly under the "Execute the following transaction out-of-band" header is the natural home, since it explains WHY the user is reading this error at all. Keep the absent-field-means-silence rule exactly as is.

## Also fix while you are there, both cheap

- **The `not-a-candidate` wording is slightly wrong.** Candidacy requires the resolved signer to be `remote` as well as the address being a named account absent from `eth_accounts`: `needsImpersonationForRun` excludes `signerOnly` and `wallet`. "Only NAMED accounts absent from `eth_accounts`" is broader than the real rule. It is unreachable at the seam (a local signer classifies as `local` and never gets there), so this is text-only, in both the message wording and the `documentation.md` sentence. Make it true anyway; this repo has paid for plausible-but-wrong statements more than once.
- **Record the decisions.** No `## Decisions` block and no `decisions-*.md` landed, unlike every sibling task in this chain, so the two judgement calls below had to be reconstructed from source by the review. One note is enough.

## Ratified, do NOT churn

- **The typed `autoImpersonation: 'attempted' | 'not-a-candidate'` field on `UnknownSignerErrorData`.** The task asked only that the message say so, and this adds published API surface to `@rocketh/core`, which `AGENTS.md` lists under "ask first". I am ratifying it: letting a consumer branch programmatically instead of regexing a message is genuinely better, and it is additive and optional. Accepted cost, which is why it belongs in the decisions note: a third variant later widens a published union.
- **Inferring `attempted` from the candidate set at the call site** rather than having the helper report back. Accurate today because the helper loops every address with a per-address try/catch. It is a second source of truth, so note in the decisions record that if the helper ever early-returns or filters, this claim goes stale.

## Verified

Every criterion has a real test: attempted-and-unsupported, attempted-and-refused, never-attempted, unchanged-when-off (no new noise on the common path), the deployment funnel, signability classification re-pinned unchanged, and both an impersonated and a local account still broadcasting untouched while impersonation is on and unsupported. The three documentation constraints are all present and each is a verified fact about the code. Lockfile: zero diff.

wighawag:
Approved.\n\nAll three Gate-3 blockers are fixed in the bytes, not just claimed. The note is now pushed at line index 2, directly under the out-of-band header and above contract:/from:/to:/value:/data:, and errors.test.ts pins that index against a 4000-byte data field so a future reorder that re-buries it fails. The not-a-candidate wording was narrowed in both the message and documentation.md constraint 1 to the real rule (named AND remote-signer AND absent from eth_accounts); the only surviving mention of the old broader phrasing is the JSDoc, which states all three conjuncts correctly. A decisions note landed with four entries including the accepted costs the requeue asked for.\n\nClaim-check: verified against source, not the task prose. impersonateAccounts loops every address with a per-address try/catch and no early return, so inferring attempted from the candidate set is accurate. from is lowercased at the seam and matches the lowercased set. ExecutionParams really carries provider and autoImpersonate at top level, {default: '0x...'} is a valid AccountType, and @rocketh/web never supplies promptExecutor, so the doc claim that ask degrades to throw there holds.\n\nCleanup-vs-behaviour: purely additive. The field is optional and absent when autoImpersonate is off, and a test asserts byte-identity with a freshly built error on that path. Classification, policy placement and broadcast are re-pinned by the second describe block.\n\nCoherence: capability-before-seam vs policy-after stays intact (ADR 0006); the outcome is read only where the error is built, inside the unsignable branch, and cannot reach the policy. No 'impersonate' policy value is implied, and the docs say so.\n\nDestination check: merged as written, both silences the task named are closed. A user who enabled impersonation against the wrong node is told which of the two things happened, the common path is unchanged, and the browser/fork route is documented with its three real constraints in the section the rest of the chain built. Four non-blocking ratification/wording findings recorded.

wighawag:
APPROVE ✅ — Gate-3 re-review (my BLOCK is resolved)

- **Placement fixed.** The note now renders before `contract:`/`from:`/`to:`/`value:`/`data:`, so a deployment's creation bytecode can no longer bury it. The absent-field-means-silence rule is untouched.
- **Wording is now accurate**, and better than what I asked for: "the candidates are the NAMED accounts (declared in the `accounts` config) **that the node would otherwise have to sign for** … an unnamed account, a bare `from`, or an account that already has its own signer is never impersonated." That captures the `remote`-signer half of `needsImpersonationForRun` in plain language, without making the reader learn the predicate's name.
- **Decisions note landed**, so the typed-field and inferred-`attempted` judgement calls are now discoverable rather than reconstructable.
- The `attempted` message names the fix ("only a fork or dev node, such as anvil or hardhat, implements that RPC"), which is the difference between telling the user what happened and telling them what to do about it.

Re-verified on the updated tip: lockfile zero diff, and the ratified decisions were not churned.

Everything from my first pass stands: purely additive so control flow is provably unchanged, all eight tests map to criteria (including the never-attempted shape, the no-new-noise-when-off case, the deployment funnel, the re-pinned signability classification, and both an impersonated and a local account still broadcasting untouched), and the three documentation constraints are each verified facts about the code.

Merging. This completes the `unknown-signer-interactive` chain.

PR #103, opened 2026-08-27
feat(execute-guard-seam-and-call-kind): The execute guard seam: a decla…

wighawag:
Task: `work/tasks/done/execute-guard-seam-and-call-kind.md`

All green. Here is where things stand.

## What I found

This was a continuation. The prior attempt's implementation had already landed on `work/task-execute-guard-seam-and-call-kind`, and the requeue was **not** a code failure: commit `b64b3837` on `main` records the acceptance gate dying on its very first step, `pnpm format:check`, over three `*asterisk*` emphasis spans in the transcribed `## Decisions` block of the done record. Prettier normalises markdown emphasis to `_`. Build, typecheck and test never ran. A follow-up commit on the branch (`4861e234`) already normalised those three spans.

So my job was to verify the prior work rather than restart it, and to make sure the same formatting trap cannot re-fire on this attempt's transcription.

## Verification I actually performed

- **Drift check against reality.** `docs/adr/0013-the-execute-guard-is-a-declared-read.md` exists and matches what was built: declared read not opaque predicate, `kind` discriminated from the first commit, `satisfied` as the primary verdict, guard evaluated before the transaction is built so it reaches neither the broadcast choke point nor the unknown-signer seam, nothing persisted, unguarded return type untouched. No stale premise found.
- **`read.ts` is genuinely a verbatim move**, not a rewrite. I diffed `packages/rocketh-read-execute/src/read.ts` line-by-line against `origin/main:packages/rocketh-read-execute/src/index.ts` lines 224-365: identical except one trailing blank line. The `ReadFunction` / `ReadFunctionByName` / `ReadingArgs` types moved unchanged too.
- **The type-level assertions are really enforced.** `tsc --noEmit --listFiles` in the package confirms `test/execute-guard.integration.test.ts` is in the checking program, so the `IsExactly<>` pins and the four `@ts-expect-error` directives are checked by `pnpm typecheck`, per the CONTEXT.md convention.
- **The full gate, in the runner's order.** `format:check`, `sync:template:check`, `changeset status --since=main` (`@rocketh/read-execute` minor + three patch dependents), `build`, `typecheck`, `test` (83 files, 983 tests), `test:getting-started` (exit 0). All green.
  - One trap worth naming: `pnpm typecheck` hung and then reported bogus `Cannot find module '../generated/artifacts/index.js'` errors in every demo, purely because `pnpm build` was a 100% Nx cache hit that never regenerated the demoes' `generated/` trees. After `pnpm build --skip-nx-cache` the whole workspace typechecks clean. That is a local-cache artefact, not a repo problem, so I did not file a note.

## The one change I made

`packages/rocketh-read-execute/README.md`: the `tx` entry claimed "Returns the transaction hash", while `tx` resolves to a receipt (`env.broadcastExecution` is declared `Promise<EIP1193TransactionReceipt>` in `packages/rocketh-core/src/types.ts:781`). That is the exact same false claim this task asked me to correct in the `TxFunction` type alias, left standing in the prose one file over. Corrected.

## Acceptance criteria

Every box is satisfied: guard accepted on `execute`/`executeByName` discriminated on `kind`; reads another target and defaults to the executed contract; typed against the ABI it reads (pinned by `@ts-expect-error`); a satisfied guard broadcasts nothing (asserted from recorded provider requests, not the return value); unsatisfied executes byte-identically to unguarded; unguarded return type pinned exactly unchanged; `evaluateGuard` exported curried and callable standalone; evaluation returned on both paths; the `TimelockController` `OperationState` negation exists as a fixture; the three stale type aliases corrected; changeset present.

Working tree contains only the README edit.

## Decisions

- **Option name `guard`, member `kind: 'call'`, verdict `satisfied`.** Fixed by the spec and ADR 0013; recorded only for completeness. `equals`, output selection and the `storage` kind are deliberately absent, they are the sibling tasks.
- **Result discriminant is `outcome: 'skipped' | 'sent'`, types `GuardedExecutionResult = SkippedExecution | SentExecution`.** `outcome` rather than `status`, because `status` already means the receipt's success flag one level down (`result.receipt.status`), and two `status` fields on nested objects meaning different things is exactly the re-meaning the coherence check warns about. `skipped` is not named after a transaction, per the task. Touches: `execute-guard-storage-kind`, `-equals-and-output-selection`, `-evaluation-is-legible` and `-convergence-scenario` all consume these names.
- **The evaluator lives in a new `src/guard.ts`, and `read` moved to a new `src/read.ts`.** The evaluator must go _through_ `read`, and `execute` (in `index.ts`) must call the evaluator, so leaving `read` in `index.ts` would have made `index` and `guard` a cycle. `read.ts` is a verbatim move, verified line-by-line against `origin/main` (identical but for one blank line), and the root still exports `read`/`readByName` identically.
- **`evaluateGuard(env)(guard, defaultTarget?)` throws when the guard names no `on` and no default target is given.** A NEW refusal, reachable only from standalone evaluation, since `execute` always passes the executed deployment. Alternative considered: make `on` mandatory in the type, rejected because "defaults to the contract being executed" is an acceptance criterion.
- **The unguarded call signature is `ExecutionArgs & {guard?: undefined}`.** This accepts everything the pre-guard signature accepted (the return type is pinned as exactly unchanged), but it makes a possibly-present guard (`guard: maybeGuard`) match neither signature, i.e. a compile error, rather than silently typing as an unguarded call whose receipt a skipped result would not have. Touches user-visible ergonomics: a caller who conditionally attaches a guard must branch instead.
- **The guard reads at `latest` only.** It takes `on`, `functionName`, `args` and `satisfied`, and deliberately not `read`'s `blockTag` / `blockNumber` / `account`. Additive later if a real case appears.
- **The guard is evaluated before `env.resolveAccount(account)`.** A satisfied guard costs one read and nothing else, and cannot fail on account resolution for work that is already done.
- **README: a guard section, plus two corrected sentences.** "The four functions" (which this change made wrong) and the `tx` entry, which claimed a transaction hash while `tx` resolves to a receipt: the same false claim the task asked me to fix in `TxFunction`.
- **Decisions text uses no `*`-emphasis.** The previous attempt's gate went red at `pnpm format:check` on the transcribed done record, because Prettier normalises markdown emphasis to `_`. I ran this exact block through `prettier --check` with the repo config before emitting it.


wighawag:

Every acceptance criterion is met, and the two that are easiest to fake are genuinely done.

| Criterion | Evidence |
| --- | --- |
| `guard` on `execute` + `executeByName`, discriminated on `kind`, `call` implemented | `src/guard.ts`, `ExecuteGuard` is a union of one with `kind` present from the first commit |
| Reads a target OTHER than the executed contract; defaults to it | tests "reads a contract OTHER than the one being executed" and "defaults its target to the contract being executed" |
| Typed against the ABI it READS | two `@ts-expect-error` pins: a getter absent from the read ABI, and a `bytes32` arg passed as a number |
| A satisfied guard sends NO transaction | asserted from `provider.getRequests()` filtered to `eth_sendTransaction` / `eth_sendRawTransaction`, not from the return value |
| An unsatisfied guard executes exactly as today | calldata compared byte for byte against the transaction the unguarded form produces |
| Unguarded return type UNCHANGED | `IsExactly<typeof result, EIP1193TransactionReceipt>` on both `execute` and `executeByName` |
| Evaluator exported, callable standalone | `evaluateGuard(env)(guard, defaultTarget?)`, curried, plus its own tests including the refusal when there is no target at all |
| Evaluation returned on both paths | test "reports the evaluation on BOTH paths" |
| Negation expressible, with the timelock enum | `satisfied: (state) => state === OperationState.Done` at runtime and in a type pin |
| Stale aliases corrected | `ExecuteFunction`, `ExecuteFunctionByName` and `TxFunction` now describe what their functions actually return |

Three things worth calling out beyond the checklist.

**The conditional signature is better than the task asked for.** `ExecutionArgs & {guard?: undefined}` makes a POSSIBLY-present guard (`guard: maybeGuard`) match neither overload, so it is a compile error rather than being typed as an unguarded call whose receipt a skipped result would not have. That is a real unsoundness closed, and it is recorded in the Decisions block rather than left for a reviewer to find.

**The read extraction is justified, not incidental.** `read` moved to `src/read.ts` because the evaluator must go through it while `execute` must call the evaluator, which would have made `index.ts` and `guard.ts` mutually dependent. Verbatim move, root exports unchanged.

**Off-path finding captured, not fixed:** `execute` forwards `gas: 0n` as a raw bigint because that one field uses `&&` where every sibling uses a ternary. Correctly written up as an observation instead of expanding the task.

The ADR-0013 constraint that motivated the whole shape is honoured: the guard is a declared read, the evaluation is a value on both paths, and `guard.ts`'s module comment says why an opaque predicate was refused.

Gate-1 ran the full verify on the rebased tip this time (format, changeset status, build, typecheck, test, test:getting-started) after the earlier bounce, which was cosmetic and is written up in `work/notes/observations/decisions-block-formatting-reds-the-gate-after-a-green-build.md`.

PR #105, opened 2026-08-27
feat(execute-guard-equals-and-output-selection): equals sugar, selectin…

wighawag:
Task: `work/tasks/done/execute-guard-equals-and-output-selection.md`

Clean. Here is my report.

## What I did

This was a **continuation**. I first checked the task against reality and reviewed the prior attempt (`git diff origin/main...work/task-execute-guard-equals-and-output-selection`).

**Drift check (no drift, no double-build).** The seam task `execute-guard-seam-and-call-kind` landed (#103) with exactly the shape this task assumes: `CallGuard` discriminated on `kind: 'call'`, `evaluateGuard(env)` curried, and a `CallGuardEvaluation` record. It did **not** solve comparison: its only verdict form was `satisfied`, and the value reached it raw. So this task was still real work, not a re-build.

**The prior attempt was substantively complete and its bounce was environmental**, not a defect in the branch. The acceptance gate failed on a fresh worktree because nx reported 23/23 build cache hits, so the demos' `hardhat compile` never ran and their gitignored `generated/artifacts` was never produced, after which `tsc --noEmit` failed with TS2307 in files this task never touched. That is already captured on `main` as `work/notes/observations/a-cached-demo-build-leaves-no-generated-artifacts-so-typecheck-fails.md`; I did not expand scope to fix it, and I ran the build here with `NX_SKIP_NX_CACHE=true` (a workaround, matching the runner's re-dispatch).

**I verified rather than trusted the two claims the task flags:**

- `AccessManager.hasRole` return shape, fetched from OpenZeppelin's `contracts/access/manager/IAccessManager.sol` at `master` (header: last updated v5.7.0): `function hasRole(uint64 roleId, address account) external view returns (bool isMember, uint32 executionDelay)`. The fixture in the test matches it exactly, so the tuple topology test stands as written.
- viem's decode shape, which the selection feature rests on: one output unwraps, several become an array, none is `undefined`. `abi-comparison.ts` and `selectOutput` are both keyed off that, and the single-output-selection test pins it.

**What I added.** One gap: "arrays and struct returns compare deeply, with the same per-type rule applied elementwise" was implemented in `valuesEqualForAbiType` (the `T[]` and `tuple` branches) but pinned by no test, so nothing could have caught a regression there. I added two tests to `packages/rocketh-read-execute/test/execute-guard-equals.integration.test.ts`, plus the two ABI fixtures they read against (`getFlashBorrowers() returns (address[])` and `getReserveConfig() returns (tuple(address treasury, string label, uint256 cap))`):

- an `address[]` return matches a lowercased expected list against the checksummed decoded one, per element, and still refuses a different member and a shorter list;
- a struct return judges each component under **its own** declared ABI type: the `address` folds case, the `string` does not, the `uint256` does not coerce against a number. That is the case that proves the rule is per-component and not per-value, since two of the three fields arrive as JavaScript strings.

I confirmed both are falsifiable: with the `address`/`bytesN` folding branch disabled, they fail (alongside the three existing folding tests) rather than passing vacuously. Source restored afterwards; `git status` shows only the test file modified.

## Acceptance criteria

All met. `equals` skips/executes correctly; `output` selects by name and by position and is typed against the ABI outputs (`@ts-expect-error` pins on `'isMemebr'` and on position `2`); the selected value reaches both `equals` and `satisfied`; address casing matches in both directions (the reverse direction pinned directly against `valuesEqualForAbiType`, because viem's decoder always hands back the checksummed form so only one direction is reachable through a `call` guard); an upper-case `bytes32` matches a lower-case read; two `string`s differing only in case do not match; a bigint does not match a number; a multi-output return compares elementwise, and now arrays and structs do too; the evaluation reports `value`, `selected` and `expected`, present only when they mean something. A changeset is present (`.changeset/execute-guard-equals-and-output-selection.md`, `@rocketh/read-execute: minor`).

## Checks

`pnpm format:check`, `pnpm sync:template:check`, `pnpm changeset status --since=main`, `pnpm build` (cache skipped), `pnpm typecheck`, `pnpm test` (84 files, **998 tests**), `pnpm test:getting-started`: all green. Working tree carries only the intended test change; the `my-rocketh-project/` tree that `test:getting-started` creates is already in `.gitignore`.

## Decisions

The first seven are the prior attempt's, restated because they are the operative rationale for the code that is landing; the last is mine.

- **The selector is spelled `output`, and takes a name or a position (`output: 'isMember'` / `output: 1`).** `output` is the ABI's own word for the thing being named (`outputs` in the ABI item), so it does not re-mean anything in `CONTEXT.md` or invent a parallel vocabulary. Alternatives considered: `select` (a verb in a bag of nouns, and it invites a path language later), and `pick`/`at` (neither says _of what_). It is typed as a union over the declared outputs, so `output: 'isMemebr'` and `output: 2` on a two-output function are compile errors. Touches: `execute-guard-storage-kind` (a slot has no declared outputs, so a storage guard should carry no `output` at all rather than a degenerate one), and `execute-guard-evaluation-is-legible`, which renders the field.
- **A selection on a SINGLE-output function is accepted, and is the identity.** viem unwraps a lone output before the guard sees it, so selection is only _meaningful_ when there are several. `output: 0` (or the output's name) on a one-output function returns that same value rather than being a compile error. Reason: refusing it would need a second, subtler type-level rule whose only effect is to punish an author for spelling out what they are asserting, and the runtime meaning is unambiguous. Alternative considered: make the selector type `never` when there is exactly one output. Touches nothing outside this module; a test documents the behaviour so it cannot be "fixed" by accident.
- **Case folds for `address` and for EVERY hex-bytes type (`bytes`, `bytes1` … `bytes32`), not just `bytes32`.** The spec and ADR 0013 name `bytes32` because that is what a governance guard reads (role ids, salts, operation ids), but the _argument_ is that a hex byte string's casing carries no meaning, and that is equally true of a `bytes4` selector. Folding `bytes32` while comparing `bytes4` case-sensitively would be a rule nobody could predict. Alternative considered: fold exactly `address` and `bytes32` as literally written. Touches `execute-guard-storage-kind`, whose declared word types (`address`, `bytes32`, `uint256`, `bool`) should route through this same module rather than grow a second comparison vocabulary.
- **`equals` and `satisfied` are MUTUALLY EXCLUSIVE, and stating neither throws.** A guard states its verdict once. Stating both is a compile error (each is `?: never` in the other's variant); stating neither is a runtime error naming the guard's function, reachable only by defeating the types. `equals` is not a shorthand that _also_ synthesises a `satisfied`, and neither silently wins over the other, because a guard whose verdict is ambiguous is exactly the double-execution hazard the guard exists to remove (ADR 0012). This is a NEW refusal, which is why it is recorded here.
- **A bigint against a number of the same magnitude is "not satisfied", NOT an error.** The type already refuses it; at runtime (a value out of JSON, an `any` at the call site) the guard reports unsatisfied and the call proceeds, rather than throwing. Alternative considered: throw, on the grounds that a guard that cannot produce a meaningful verdict should be fatal (ADR 0013's rule for a guard that _throws_). That rule is about the READ failing, not about a comparison legitimately coming out false, and "proceed" is the safe direction here: it re-sends a call rather than skipping one that was needed.
- **The evaluation carries `output`, `selected` and `expected`, PRESENT only when they mean something.** No `undefined`-valued keys, so a record never claims a selection that was not made or an expected value that was never stated, and `'selected' in evaluation` is a usable question. `expected` is the name for the `equals` operand ("what it was compared against"); `value` keeps its existing meaning, the WHOLE decoded return, so the seam task's tests and any consumer of it are untouched. Touches `execute-guard-evaluation-is-legible` and `state-change-provenance`, which both read this record.
- **The comparison lives in its own module, `src/abi-comparison.ts`, and is NOT on the package root.** The root surface may hold only curried `(env) => …` functions (`AGENTS.md`, `CONTEXT.md`), and this is a pure function of an ABI parameter and two values, so exporting it there would be refused by name at deploy-script run time. It stays internal, which also keeps ADR 0013's closing consequence true: `satisfied` receives the value RAW and there is no second, user-facing comparison vocabulary competing with `equals`. The test imports it directly from `../src/abi-comparison.js` to pin the symmetric address case.
- **A struct's components are read by NAME or by POSITION, whichever the value offers, on both sides independently.** viem decodes a tuple as an object when every component is named and as an array otherwise, and an author writing an `equals` for a struct will naturally reach for the object spelling. Rather than normalise one shape into the other, `component()` accepts either on either side, so an object expected value compares against an array-decoded actual and vice versa. Alternative considered: require both sides to be the same shape and return `false` otherwise, which would have made the guard's verdict depend on whether the Solidity author happened to name every field, a detail invisible from the deploy script. This is a behaviour of `equals` only, and the new struct test is what stops it from being narrowed by accident; nothing outside `abi-comparison.ts` is touched.


wighawag:
APPROVE ✅ — Gate-3

All criteria met. The three that could have been faked, and were not:

- **The case-SENSITIVE string assertion exists** ("does NOT match two strings differing only in case"). That is the one a naive lowercase-everything implementation passes wrongly.
- **The selector is typed against the ABI outputs**: `output: 'isMemebr'` and `output: 2` on a two-output function are both `@ts-expect-error` pins, not prose.
- **The OZ return shape was actually verified**, which the task explicitly required rather than trusting: `hasRole(uint64,address) returns (bool isMember, uint32 executionDelay)`, cited to `AccessManager.sol` with the version and date. This was the one deferred verification I flagged when tasking, and it is now closed.

Two in-scope decisions I looked at closely and agree with:

- **Case folding was widened from `bytes32` to every hex-bytes type.** The ADR names `bytes32` because that is what governance guards read, but the argument is about hex casing carrying no meaning, which is equally true of a `bytes4` selector. Folding one and not the other would be unpredictable.
- **A bigint against a number of the same magnitude reports unsatisfied rather than throwing.** Correct direction: ADR 0013 makes a failed READ fatal, not a comparison that legitimately comes out false, and unsatisfied re-sends a call rather than skipping one that was needed.

`abi-comparison.ts` correctly stays off the package root, since the root may hold only curried `(env) => …` entries.

Gate-1 green. The earlier bounce on this branch was environmental (an nx cache hit left the demos without their generated artifacts, so typecheck failed in files this diff never touches): written up in `work/notes/observations/a-cached-demo-build-leaves-no-generated-artifacts-so-typecheck-fails.md`.

PR #106, opened 2026-08-27
feat(execute-guard-storage-kind): The storage guard kind: read a slot,…

wighawag:


wighawag:
APPROVE ✅ — Gate-3

Both required topologies are present and are the real ones: the ProxyAdmin case (call the admin, read the EIP-1967 implementation slot on the PROXY) and the Aave case (call the registry whose own `getPool()` cannot tell, read the slot on the proxy behind it). A call-only implementation fails both, which is why the kind exists.

Three decisions I checked rather than took on trust:

- **`as` is required and its members are ABI TYPE names**, so the declared interpretation is handed straight to the comparison module as `{type: guard.as}`. That makes "an address in a slot folds case exactly as an address from a getter" true by construction instead of by a second implementation of the same rule. Correct call.
- **A word that does not fit its declaration throws, with `address` as the deliberate exception.** A `bool` slot holding `2` is almost always a PACKED slot, which no whole-word read can answer, so guessing `non-zero means true` would be exactly the guess ADR 0013 forbids. `address` ignoring the high bytes matches `@rocketh/proxy` and every EIP-1967 reader.
- **The three exported result aliases were reparametrised by the evaluation type**, which is breaking. I verified the claim behind the `minor`: nothing outside `@rocketh/read-execute` references them, and neither earlier changeset has been released, so they have never shipped.

Note on the gate: Gate-1 was run BY HAND rather than by the runner, and nothing was skipped. I ran all six steps of the configured verify on this exact tree, unloaded: `format:check`, `sync:template:check`, `changeset status`, `build`, `typecheck`, `test` (85 files, 1010 tests) and `test:getting-started`, all green. The runner's own gate had bounced this branch twice on `@rocketh/export` tests that compile TypeScript in-process against a 5s default timeout and take ~950ms each when the machine is idle: a load flake in a package this diff never touches, written up in `work/notes/observations/export-typecheck-tests-flake-under-load-on-the-default-5s-timeout.md`. The one gap to be honest about: I verified the branch tip, and main has moved only by documentation commits since.

PR #107, opened 2026-08-27
feat(execute-guard-failure-is-fatal): A guard that throws fails the run…

wighawag:
Task: `work/tasks/done/execute-guard-failure-is-fatal.md`

I reviewed the prior attempt's work, verified it against current reality, and ran the full acceptance gate.

## What I found

The prior attempt on `work/task-execute-guard-failure-is-fatal` (commit `c147c759`) had already implemented the task completely. Its requeue was **not** caused by its own change: main's `1535d711` records the gate failing at its very first link on a file this task never touched — `work/questions/task-execute-guard-storage-kind.md`, a runner-generated question sidecar that is not prettier-clean. Main has since cleared it (`1a819fee`), and this branch already contains that commit.

## Drift check (as the prompt asked)

- Both guard kinds landed: `execute-guard-seam-and-call-kind` (#103) and `execute-guard-storage-kind` (#106) are on main.
- The evaluator has exactly **one** place failures can be caught: `evaluateGuard` in `packages/rocketh-read-execute/src/guard.ts` is the sole entry point; `evaluateCallGuard` / `evaluateStorageGuard` are module-private and unreachable otherwise (verified by grep across `packages/`).
- The earlier tasks did **not** already make failure fatal in a discriminating way — they left "nothing here catches" as a comment with no error identity and no test that would go red on a fall-through. This task's change is what supplies both.

## What the change on this branch does (reviewed, endorsed, unchanged)

- `packages/rocketh-read-execute/src/errors.ts` (new): `GuardEvaluationError` + `GuardEvaluationErrorData` + `describeGuard`, on the `./errors` subpath so the root export surface stays function-only for `withEnvironment`.
- `packages/rocketh-read-execute/src/guard.ts`: one `try`/`catch` in `evaluateGuard`, with both branches **awaited inside** the `try` (returning the promise would put every asynchronous failure outside the catch). Foreign errors wrapped with `cause` preserved and quoted; already-ours rethrown. The asymmetry comment sits at the choice site, and at the `execute` call site.
- `packages/rocketh-read-execute/test/execute-guard-failure.integration.test.ts` (new, 550 lines): revert, empty return after exhausted retry, unknown address, throwing `satisfied`, refused `eth_getStorageAt`, undecodable word, plus the regression pin (identical script, "not satisfied" broadcasts once / "cannot say" broadcasts nothing). All "nothing executed" assertions read the recorded provider requests. The retry count is read off `env.context.retry.maxRetries` rather than hardcoded, so tuning the policy re-tunes the assertion instead of invalidating it — I checked `read.ts` and the loop does issue `maxRetries + 1` `eth_call`s and rethrows the original `AbiDecodingZeroDataError` when exhausted, so the test is about the end state as required.

## Gate, run locally on the rebased tip

| step | result |
| --- | --- |
| `pnpm format:check` | green (the stale sidecar that bounced it is gone) |
| `pnpm sync:template:check` | green |
| `pnpm changeset status --since=main` | green (`@rocketh/read-execute` minor) |
| `pnpm build` | green (23 projects) |
| `pnpm typecheck` | green |
| `pnpm test` | 1022/1023; the single red is the known `@rocketh/export` tsc-in-process 5s-timeout flake, already documented in `work/notes/observations/export-typecheck-tests-flake-under-load-on-the-default-5s-timeout.md`. Re-run in isolation: 35/35 pass. |
| `pnpm test:getting-started` | green |
| `packages/rocketh-read-execute` alone | 82/82 pass |

I also confirmed the `./errors` subpath resolves from the built output (`import('@rocketh/read-execute/errors')` yields `GuardEvaluationError`, `describeGuard`).

Working tree is clean; no stray untracked files.

I made **no new** source changes this cycle and **no new** decisions. The block below is the prior attempt's rationale for the change this branch carries; I reviewed it and endorse it. Note for the runner/human: it is **already present verbatim** in `work/tasks/done/execute-guard-failure-is-fatal.md` (the prior attempt's commit did the done-move), so it should be replaced rather than appended a second time.

## Decisions

**The error type is a new `GuardEvaluationError`, on a `./errors` subpath.** A plain `Error` cannot be identified by a caller and carries no structured `target`, and the criterion asks the surfaced error to _identify_ the guard. It lives on `@rocketh/read-execute/errors` rather than the package root because `withEnvironment` calls every root export as `value(env)` and refuses a class by name, exactly the reasoning `@rocketh/unknown-signer` records for `UnknownSignerError`. Alternatives considered: reuse a plain `Error` (rejected, nothing to catch on); put it in `@rocketh/core` beside `UnknownSignerError` (rejected, the guard is a `read-execute` concept and core should not learn about it); export it from the root (rejected, breaks the documented extension-spread idiom). What it touches: a new `./errors` entry in `packages/rocketh-read-execute/package.json`, the README, and any later `document-execute-state-guard` or collector work that renders a guard failure. Its _type_ (`GuardEvaluationErrorData`) is re-exported from the root, since types erase and cost the spread nothing.

**Foreign failures are WRAPPED, not rethrown; our own are rethrown untouched.** A bare `AbiDecodingZeroDataError`, or a node's `method eth_getStorageAt is not available`, does not say which of a script's guards asked or which contract it read, and those are the two facts needed to fix the script. So the underlying error is kept whole on `cause` _and_ quoted in the message (`the guard on slot 0x36… of 0xee… could not be evaluated, so nothing was executed: <underlying>`), which is how "preserves the underlying failure rather than replacing it" is satisfied on both channels. Alternatives considered: rethrow unchanged (rejected, loses the guard's identity, which is the criterion); replace the message with a guard-specific one (rejected, loses the actionable cause); attach the guard to the original error object by mutation (rejected, mutating someone else's error is worse than a `cause` chain). An error that is _already_ a `GuardEvaluationError` is rethrown as is, so the message never says the same thing twice.

**The guard module's own declaration throws became `GuardEvaluationError` too** (no target, no verdict, an `output` the ABI does not declare). They are declaration mistakes rather than chain failures, but the consequence has to be identical, and one error type means a caller writing `instanceof` covers every way a guard fails to answer. Side effect worth naming: a call guard's "states no verdict" and "selects the output" messages now also carry ` of <target>`, which they did not before; the existing tests' `/no target/i` and `/bool/i` assertions still match, since the underlying wording is preserved. This touches the three earlier `execute-guard-*` tasks' error strings and nothing else.

**No new ADR.** ADR 0013's consequences already state "a guard that reverts or otherwise throws is FATAL, never _not satisfied_"; this task implements that decision rather than making a new one, so the rationale went to the choice site and to `errors.ts` instead of to `docs/adr/`.


wighawag:
APPROVE ✅ — Gate-3

The criterion that mattered here is the one that is easy to satisfy vacuously, and it is done properly: "sends when the guard says not yet, and sends NOTHING when the guard cannot say" runs the identical script and identical guard twice, varying only the chain's answer, so a regression to `catch { satisfied: false }` anywhere on the evaluation path makes it broadcast the transaction the guard exists to withhold. That is a real mutation pin, not a presence check.

The retry boundary is tested from both sides (fatal once exhausted, NOT fatal while a retry can still succeed), which is what stops a later tuning of the retry policy from turning a fatal guard into a fall-through without anything going red.

Two decisions checked:

- **`GuardEvaluationError` on a `./errors` subpath, not the package root.** Correct, and for the documented reason: `withEnvironment` calls every root export as `value(env)` and refuses a class by name at deploy-script run time, which is exactly why `@rocketh/unknown-signer` keeps `UnknownSignerError` on `./errors`. The error's TYPE is re-exported from the root, which is free since types erase.
- **Foreign failures wrapped with `cause` preserved and quoted; our own rethrown untouched.** Satisfies "preserves the underlying failure rather than replacing it" on both channels, and avoids the message repeating itself.

Gate-1 green through the full chain. This branch was bounced once by a stale sidecar on main rather than by anything in the diff: a bounce sidecar quotes raw gate output in a blockquote, prettier reflows it, and `format:check` is the first link of the gate, so one bounce reds the next task and writes another dirty sidecar. Written up in `work/notes/observations/a-bounce-sidecar-is-prettier-dirty-so-it-bounces-the-next-task.md`.

PR #108, opened 2026-08-27
feat(execute-guard-convergence-scenario): The loop closes: a deferred S…

wighawag:


wighawag:
APPROVE ✅ — Gate-3

This is the story the whole feature exists for, and it reads as documentation rather than as coverage. Run 1 defers and hands over the exact transaction, the Safe executes it out of band (modelled by moving the EIP-1967 slot, which is all that transaction does), run 2 runs the SAME script unedited, the guard reads the new state and the step is skipped. `catchUnknownSigner` returns null on run 2 because there is nothing left to catch.

Four things I checked specifically:

- **The skip is earned**, not incidental: a sibling test defers AGAIN when the Safe has not acted, so run 2's skip is caused by the chain state and not by anything about being a second run.
- **The mixed run** is demonstrated, with signable steps broadcasting on both runs while the Safe-only step defers then skips.
- **Nothing is persisted**: no file appears and the deployment records are unchanged across the runs, which is the guarantee ADR 0012 rests on.
- **The test is honest about what its evidence can prove.** It notes that a deferral costs NO RPC, since the seam throws before the transaction leaves rocketh, so run 1 and run 2 have the same single read in the request log. It therefore proves "the transaction was never even built" with a separate test that makes the Safe signable. That is a distinction most tests would have got wrong while still passing.

The evaluation assertion is a nice incidental proof of the storage kind's design: `value` is checksummed and `expected` is not, they differ in spelling, and they match because the declared interpretation is what folds case.

Gate-1: run by hand, unloaded, all steps green except the known `@rocketh/export` flake (1028 of 1029 tests passed; that file then passed all 35 in isolation). That flake has now bounced three gates in this drive and is written up, with the sharpening that its trigger is the repo-wide `pnpm test` itself rather than extraneous load.

PR #109, opened 2026-08-27
feat(execute-guard-evaluation-is-legible): A skipped step explains itse…

wighawag:
Task: `work/tasks/done/execute-guard-evaluation-is-legible.md`

Done. Everything is green.

## What I did

**Drift check first.** The task's premises hold: `GuardEvaluation` (both `CallGuardEvaluation` and `StorageGuardEvaluation`) exists in `packages/rocketh-read-execute/src/guard.ts` with `target` / `functionName` / `args` / `value` / `output` / `selected` / `expected` / `satisfied` (call) and `target` / `slot` / `word` / `as` / `value` / `expected` / `satisfied` (storage); both guard kinds and the `output` selection landed; and nothing in `@rocketh/read-execute` reported anything to the user yet (`grep showMessage` over the package: no hits before this change). So this is a first message, not a second.

**Test first** (`packages/rocketh-read-execute/test/execute-guard-message.integration.test.ts`, 12 cases, red before the code, green after): exact-string assertions on what the user is told for a call guard (with/without args, selection by name and by position, `satisfied` predicate), a storage guard (decoded value, non-address interpretation), `executeByName`, silence on the sent path and on unguarded calls, one line + no em dash, the standalone `evaluateGuard` saying nothing, and a case pinning that the line's content is the returned evaluation record's content.

**Production code**:
- new `packages/rocketh-read-execute/src/guard-message.ts` with `describeSkippedStep(functionName, evaluation)`, rendering entirely off the evaluation record and reusing `describeGuard` from `./errors.ts` so a guard is named the same way in a skip line as in a failure message. Not exported from the package root (that surface may only hold curried `(env) => …` entries).
- `packages/rocketh-read-execute/src/index.ts`: `execute` calls `env.showMessage(...)` on the skipped path only.
- `.changeset/execute-guard-evaluation-is-legible.md` (`@rocketh/read-execute`: minor).

The lines look like:

```
skipped setPoolImpl: the guard on "hasRole" of 0xbbb…bb with args (42, 0xfff…ff) read [true,3600], output "isMember" is true, expected true
skipped upgradeAndCall: the guard on slot 0x360894…bbc of 0x1111…11 read the address 0x5FbDB2…aa3, expected 0x5FbDB2…aa3
```

**Gates**: `pnpm typecheck`, `pnpm build`, `pnpm test` (88 files / 1041 tests) and `pnpm format:check` all pass. Working tree holds only the three new/modified files plus the changeset.

## Decisions

- **Message format: ONE line per skipped step, `skipped <fn>: <guard description> [with args (…)] read <value>[, output <sel> is <selected>][, expected <value> | , accepted by its satisfied() predicate]`.** Chosen because a deploy script with dozens of guarded steps is a normal shape, and a multi-line block per step (the shape `catchUnknownSigner`'s deferral block uses) would bury the run it is meant to explain. The deferral block earns its size by being an _action item_ a human must copy into a Safe; a skip is a _fact_, so it gets a sentence. Alternative considered and rejected: the same `SEPARATOR`-fenced block as the deferral path, for visual consistency. What it touches: nothing else emits on this path today, but `state-change-provenance` and a later pending-set collector will report the same records, so the guard description deliberately goes through `describeGuard` (`./errors.ts`) rather than a second spelling.
- **The proceed path says NOTHING.** A sent call already leaves a transaction, a receipt and (when the signer is unknown) a deferral block behind it, so a line there is pure noise on the only path that has other evidence, and it would double the output of every guarded script. Alternative considered: a one-line `guard not satisfied, executing …`, which would make guard activity uniformly visible; rejected on the loudness cost, and reversible later (adding a line is additive, removing one is a regression for anyone grepping). What it touches: `@rocketh/proxy` and `@rocketh/diamond` call `execute` without guards, so neither gains output either way.
- **Hex values are quoted WHOLE, never truncated.** A shortened slot or address is not greppable and cannot be diffed against the script, which is exactly what a reader of a skip is doing. The cost is a long line (the storage case runs to roughly 250 characters). Alternative considered: eliding the middle of a `bytes32` slot; rejected because it invents a second spelling for a value that also appears in `GuardEvaluationError` messages unabbreviated.
- **A predicate verdict says `accepted by its satisfied() predicate` instead of omitting the clause.** The record legitimately has no `expected` there, and bare silence would leave a reader unsure whether an expectation existed and was dropped. What it touches: it makes the `satisfied` / `equals` distinction user-visible in output for the first time, which is consistent with ADR 0013 treating that choice as the user's, not an implementation detail.
- **The message is emitted by `execute`, not by `evaluateGuard`.** The standalone evaluator is what a later collector uses to compute the pending set _without executing anything_; a "skipped" line from there would assert a step was skipped when no step was ever attempted. What it touches: the collector work in `explore-unknown-signer-adapters` owns its own reporting and can call the same renderer if it is ever promoted out of module scope.
- **No suppression flag was added.** Making the line conditional (an option on `ExecutionArgs`, or a `log: false` mirroring `catchUnknownSigner`'s) is a new user-visible surface, and `execute` has no such switch today. Left out rather than guessed; `showMessage` is already the run's message channel, so silencing belongs there if it is ever wanted.


wighawag:
APPROVE ✅ — Gate-3

Every criterion met, and the two structural ones are done the right way round:

- **The line is read OFF the evaluation record**, never re-derived from the guard the user wrote, and there is a test pinning that the line and the returned record cannot disagree. Re-deriving would have created a second answer to "what did rocketh read", and the one shown to the user would be the one nothing ever checked against the chain.
- **It goes through `env.showMessage`**, the environment's channel, not the console (ADR 0009).

Good judgement calls on volume, both recorded: one line per skipped step regardless of kind, since a script with dozens of guarded steps is normal and a paragraph each would bury the run; and nothing at all on the path that SENDS, because that step already leaves a transaction, a receipt and possibly a deferral block behind it.

Nice details: the storage kind shows the DECODED value rather than only the raw word (the word is not what the author wrote); a predicate verdict says the predicate accepted the value instead of inventing an expected value it was never given; the read's ARGUMENTS are shown, since a guard on `hasRole(x)` is not one on `hasRole(y)`; and values are rendered whole rather than truncated, because a truncated address is not greppable. Rendering is also written so it cannot throw, which matters when the message is the only evidence of why nothing happened.

There is a test asserting one line and no em dash, which is the repo's own prose rule enforced in code.

Gate-1 green on the first attempt, no bounce.

PR #110, opened 2026-08-27
feat(document-execute-state-guard): Document the execute guard: how to…

wighawag:


wighawag:
APPROVE ✅ — Gate-3

The page documents the CODE rather than the task snapshot, which is exactly right and is where a docs task usually goes wrong. Two places it diverged from what I wrote in the task, both correctly:

- **`as` is required on a storage guard.** The spec's illustrative snippet omits it; the shipped API makes it mandatory and closed, with a type test pinning that omitting it does not compile. The page documents what shipped.
- **Trap 2 is scoped to the `call` kind.** My task said flatly "a guard against a contract that is not deployed yet aborts". That is true of a `call` guard (empty return, fatal once the retry is exhausted) and FALSE of a `storage` guard, since `eth_getStorageAt` against an address with no code answers a zero word that decodes fine and is simply "not satisfied". The merged storage tests already pin that behaviour, so the correction is right and the unqualified wording would have been wrong for half the union.

The three traps are all present, including the important one: `satisfied` hands you the value raw, so a predicate comparing addresses with `===` is wrong the moment one side is checksummed, with the correct form shown. "What the guard is not" states plainly that nothing is persisted and that there is no way to require a guard.

Honest limits, all recorded rather than hidden: the samples were type-checked once against the shipped API via a throwaway module rather than by a permanent sample test (a new checked surface this task did not ask for); the conditional cross-reference into the deferral warning was NOT made because that task has not landed; and the empty changeset carries the standing caveat that an empty changeset alone on `main` makes the release action publish nothing, silently, so it must be consumed by a real one.

Gate-1: run by hand, unloaded, all steps green, 88 test files and 1041 tests, no flake this time. The runner's own gate had bounced this branch once on the `@rocketh/export` timeout flake, which has nothing to do with a documentation diff.

PR #111, opened 2026-08-28
feat(fork-descriptor-names-the-network): Make the fork flag say WHICH n…

wighawag:
Task: `work/tasks/done/fork-descriptor-names-the-network.md`

All checks green. Here is what I did.

## Drift check

The task's premises hold against current code: `fork` was still `typeof environmentProvided !== 'string'` in `getEnvironmentName` (`packages/rocketh/src/executor/index.ts`), still carried at exactly two type sites (`ResolvedExecutionParams.environment.fork`, `Environment.network.fork`), and the only behavioural consumer is still the load-time identity skip (`packages/rocketh/src/environment/index.ts`, `loadDeployments`) plus the public field. hardhat-deploy still builds `{fork}` from `HARDHAT_FORK` and nothing else constructs a fork input. (One staging detail: the task body is at `work/tasks/backlog/fork-descriptor-names-the-network.md`, not `ready/`; the spec is in `work/specs/tasked/`. I did not move anything.)

## What changed

- `packages/rocketh-core/src/types.ts`: new `ForkInput` (`{fork, chainId?}`, what a caller passes) and `ForkDescriptor` (`{networkName, chainId?}`, what the run reports). `ExecutionParams.environment` is now `string | ForkInput`; both flag sites (`ResolvedExecutionParams.environment.fork`, `Environment.network.fork`) now carry `ForkDescriptor | undefined` and agree.
- `packages/rocketh/src/executor/index.ts`: `getEnvironmentName` builds the descriptor from the fork input alone (so a missing environment or a plain name is _not_ a fork); new exported `resolveForkDescriptor(config, executionParams)` layers the declared `environments[<name>].chain` under a caller-supplied id and stops there, never borrowing the provider's id. `resolveExecutionParams` threads the descriptor into the environment context. The `idToFetch` TODO is untouched (the sibling task owns it).
- `packages/rocketh/src/environment/index.ts`: comment at the load-time skip naming why it exists; the site itself is unchanged (still a truthiness test).
- `packages/rocketh/src/index.ts`: exports `resolveForkDescriptor` next to `getEnvironmentName`.
- Tests: new `packages/rocketh/test/fork-descriptor.test.ts` (13 cases, written before the production change: naming, both id sources and their precedence, the no-id honesty case, memory/named runs not forks, truthiness, and the discriminating pair where mainnet records load on a fork against a 31337 node but are still refused off a fork). New `packages/hardhat-deploy/test/fork-environment.test.ts` pins that the plugin still emits exactly `{fork: 'mainnet'}` and gets the descriptor plus record loading for free, with no plugin edit. Updated the two existing tests that asserted the old boolean.
- `hardhat-deploy/documentation/how-to/use-fork-testing/index.md`: one snippet interpolated `env.network.fork` into a string, which would now print `[object Object]`; changed to `env.network.fork?.networkName`. The real fork documentation is the `document-fork-runs` task's job.
- `.changeset/fork-descriptor-names-the-network.md` (`@rocketh/core`, `rocketh`: minor).

`pnpm format:check`, `pnpm sync:template:check`, `pnpm build`, `pnpm typecheck`, `pnpm test` (1056 passing) and `pnpm test:getting-started` all pass. `pnpm changeset status --since=main` still reports "no changesets were found" only because `@changesets/git` reads changesets via `git diff` against the merge base, so an _untracked_ changeset file is invisible until the runner commits; it is present and correct.

## Decisions

_The descriptor's shape and field names_ (four sibling tasks consume them). `ForkDescriptor = {networkName: string; chainId?: number}`, surfaced unchanged at `env.network.fork` and `ResolvedExecutionParams.environment.fork`. I chose `networkName` over `network` because at the path `env.network.fork.network` the word `network` already denotes an OBJECT one level up, and re-meaning it as a string there is exactly the kind of muddle `CONTEXT.md` warns about; `networkName` also says it is a NAME (the environment key under which records live). I kept `chainId` rather than `simulatedChainId`: the descriptor's whole subject is the simulated side, so the container qualifies it (`fork.chainId` reads "the forked network's chain id"), and the JSDoc pins it against `env.network.chain.id`, which stays the CONNECTED chain. Alternatives considered: `{name, chainId}` (rejected: `fork.name` reads as a name OF the fork) and a bare `{network, chainId}`. Sibling consumers get `fork.chainId` as the lookup key (`fork-semantics-come-from-the-forked-network`) and `environments[fork.networkName]` as the config entry (`fork-config-sub-key-on-the-environment`).

_The caller-supplied chain id is a sibling key on the fork input, not a nested object._ `ExecutionParams.environment` becomes `string | ForkInput` where `ForkInput = {fork: string; chainId?: number}`. Flat and additive, so hardhat-deploy's `string | {fork: string}` still type-checks and behaves identically with no edit. The alternative, `{fork: {network, chainId}}`, would have been a breaking change to the one shipped caller for no gain. The bag is where a future creation half (a block to fork from, a url) goes, per ADR 0014.

_A new exported function, `resolveForkDescriptor(config, executionParams)`._ The two id sources need the config, and `getEnvironmentName` does not take one, so I put the precedence in ONE named place rather than inline in `resolveExecutionParams` where the sibling tasks would have to re-find it. It is exported from `rocketh`'s root next to `getEnvironmentName`, which is a small public-surface addition. `getEnvironmentName` still returns a descriptor, but one that knows only what the INPUT carried; the JSDoc on both says which is which.

_A run with no environment stops taking its chain config from 31337._ This is the mechanical consequence of "the in-memory default is not a fork", and it is user-visible beyond the flag: `idToFetch = fork ? 31337 : chainId` now sends a no-environment run to the id its PROVIDER reports, so `env.network.chain.id`, the tags/policy bucket and the `deleteDeploymentsIfDifferentGenesisHash` default follow the connected node rather than being forced to 31337, and such a run now has the load-time identity check applied. In practice every such run is against a local node reporting 31337, so nothing in the repo or the getting-started flow moves. I chose this over special-casing `memory` back to 31337, which would have re-introduced the "not a named network" heuristic under a new name and contradicted ADR 0014. It touches ground the `fork-semantics-come-from-the-forked-network` and `fork-chain-identity-simulated-versus-connected` tasks will revisit, so it is named here rather than discovered there.

_The descriptor asserts no chain id when neither source said so._ Following the ADR and ADR 0012's instinct, `resolveForkDescriptor` deliberately does not fall back to the run's computed id, and the test pins that a fork of mainnet on a hardhat-shaped node (reporting 31337) reports `networkName: 'mainnet'` with `chainId` absent. The semantics lookup may use the computed id as its own fallback; the descriptor may not claim it.


wighawag:
APPROVE ✅ (conductor Gate-3: diff reviewed against the task's acceptance criteria; the acceptance gate ran green, including `test:getting-started`)

Three criteria could have been satisfied on paper here, and all three are met with tests that would actually fail if the behaviour regressed:

- **"Deployment loading still skips the chain-identity check on a fork. Tested"** is the property the whole feature exists for, and it is pinned from both sides: mainnet's records load against a node reporting 31337, *and* a non-fork run of the same environment still refuses them. The skip is scoped to forks rather than loosened.
- **"The hardhat-deploy path is unchanged. Tested"** is proved rather than read: `generateForkConfig` still constructs exactly `{fork: 'mainnet'}`, and that unchanged input yields the new descriptor with no plugin edit.
- **"Does not fabricate a chain id, does not borrow the provider's"** is tested in the case that discriminates it: nothing declared, provider reporting 31337, descriptor carries no `chainId` at all.

The truthiness contract holds (`ForkDescriptor | undefined`, so `if (env.network.fork)` reads as before), and both public type sites moved together.

Noted for the next task in this chain: `fork-descriptor.test.ts` also pins `env.network.chain.id === 31337` on a fork, which is exactly the invariant `fork-semantics-come-from-the-forked-network` must not disturb.

PR #112, opened 2026-08-28
feat(fork-semantics-come-from-the-forked-network): Split the chain-conf…

wighawag:


wighawag:

**Verification, since this landed with `--skip-verify`.** The runner's gate bounced only on the four `@rocketh/export` in-process TypeScript compile tests timing out against vitest's 5s default under the repo-wide parallel run (they took 6186ms and 6670ms here). That is a known load-sensitive flake, not this branch, so I re-ran the configured chain myself on the rebased tip in a clean scratch clone, with nothing else running. Exactly what I ran, all against commit `610beb24`, which is byte-identical to this PR's head:

- `pnpm test`: 1064/1066 passed, the only failures being 2 of those 4 export timeouts
- `packages/rocketh-export` run **alone**: 35/35 passed, including all four, which is what isolates the flake from the branch
- `pnpm test:getting-started` ✅ (the chain aborted before reaching it, so I ran it separately)

- **`env.network.chain` is UNCHANGED.** This is the one the earlier review blocked on, and the split is correctly three-way: `chainInfo` still comes from `connectedChainConfig`, so only the policy/tags consumer moved. Pinned by a test in the combination that would break, a declared simulated id of 1 against a provider reporting 31337, asserting `env.network.chain.id === 31337`. A declared simulated id reaching the transaction's `chainId` would have made a hardhat node reject locally signed transactions, and it does not.
- **The discriminating test.** Both buckets are configured with opposing values and the assertions name the forked network's, so an implementation that changed nothing would fail. Tags are asserted on the tag map deploy scripts actually read (`mainnet` true, `local` absent), not on the resolved config.
- **The connection did not move with the semantics.** Asserted positively and negatively: the provider endpoint is the local rpc url and explicitly not mainnet's.

The stale fork chain-id TODO is gone, the non-fork path collapses to the same object so it is provably unaffected, and the environment override layer still applies on top.

PR #113, opened 2026-08-28
feat(fork-chain-identity-simulated-versus-connected): A fork has two ch…

wighawag:


wighawag:

**Verification, since this landed with `--skip-verify`.** The runner's gate bounced only on the known `@rocketh/export` in-process compile tests timing out against vitest's 5s default under the repo-wide parallel run. Everything else was green there (92 files, 1078 tests). I re-ran the configured chain myself on the rebased tip in a clean scratch clone, all against commit `8a3ac509`, which is this PR's head:

- `pnpm test`: 1078/1080, the only failures being 2 of those export timeouts
- `packages/rocketh-export` run **alone**: 35/35 passed, which isolates the flake from this branch
- `pnpm test:getting-started` ✅ (run separately, as the chain aborted before it)

- **Asserted on the broadcast transaction, not on `env.network.chain`.** The integration test decodes the raw `eth_sendRawTransaction` payload and reads the `chainId` committed in the signature, under both provider shapes, with the simulated id declared as 1 in *both* runs. That is what makes it discriminating: a regression that adopted the simulated id would return 1 for the hardhat shape and fail.
- **The hardhat-deploy shape is unchanged.** Pinned at 31337 in, 31337 out, and the code path is a genuine no-op there, since `info.id` already equals the adopted id whenever the bucket was keyed by it. Only the anvil shape (node reports 1, bucket is 31337) changes, which is the bug this task exists to close.
- **The leniency is scoped to forks and nothing else was loosened.** The warning is suppressed only under `!fork`, and a non-fork genuine mismatch still warns, tested. The adoption rule is now stated at the choice site, including why `||` rather than `??` is load-bearing.

Configuration, records and semantics still follow the simulated network, so the previous task's split is undisturbed.

**Process note for the record.** The first attempt at this task died silently before its gate, with the agent's work uncommitted in the job worktree. It was recovered rather than rebuilt: the work was preserved on this branch, `requeue`d with a handoff note, and re-driven through `dorfl do`, which is why the branch carries a `wip` commit beneath the final one. The continuation is what added `@rocketh/core` to the changeset.

PR #114, opened 2026-08-28
feat(fork-aware-autoimpersonate-default): Impersonation defaults ON for…

wighawag:


wighawag:

**Verification.** This task's runner gate died on an ENOSPC in `/tmp` during the fresh-gate `pnpm install`, an environment failure that never reached the code, and the rerun bounced on the usual `@rocketh/export` timeouts. So I ran the configured chain myself on the rebased tip in a clean scratch clone:

- `pnpm test`: 1091/1093, the only failures being 2 of those export timeouts
- `packages/rocketh-export` run **alone**: 35/35 passed

- **"A fork run has auto-impersonation ON without anyone configuring it"** was the trap here, because the chain term was never `undefined`: the semantics lookup coerced an unconfigured value to `false`, so a fork-aware term appended to the chain would have been dead code. It is properly fixed at the source, by making `ChainSemantics.autoImpersonate` undefaulted while `ChainConfig` keeps its plain boolean, so no public shape changed. The test that proves it is the one with a chain entry for the forked network that simply says nothing about impersonation, which is the case a weaker implementation passes while real users get `false`.
- **A second-order version of the same trap was caught that I had missed.** The `simulatedChainId === connectedChainId` shortcut from the previous task would have re-injected the connected config's defaulted `false`, and precisely on the commonest fork there is, a hardhat node reporting 31337 for both ids. Reading through the semantics function unconditionally is the right fix and is explained at the site.
- **"An explicit `false` still wins, so the deferral path stays exercisable."** Tested at run level, at the forked network's chain config and via environment overrides, plus both directions of the precedence (run-level `true` over a forked-network `false`, and back). These can fail if the order regresses, which is what the criterion asked for. The `@rocketh/unknown-signer` scenarios pass unchanged.

The one existing assertion that was edited (`fork-semantics.test.ts`, `false` to `true`) is the genuine zero-configuration fork case, and the production-hardening page was updated rather than left stale, keeping its advice correct: a production chain is not a fork, so the default there is still off.

PR #115, opened 2026-08-28
feat(fork-config-sub-key-on-the-environment): A fork states only what d…

wighawag:


wighawag:

**Verification.** The runner's gate bounced only on the known `@rocketh/export` in-process compile timeouts, so I ran the configured chain myself on the rebased tip in a clean scratch clone:

- `pnpm test`: 1099/1102, the only failures being 3 of those export timeouts
- `packages/rocketh-export` run **alone**: 35/35 passed

- **"Declaring `whenForked` is NOT a mode switch."** The layer is read behind `fork ? environmentConfig?.whenForked : undefined`, so presence cannot imply mode, and the negative is tested from both sides: a non-fork run of an environment that HAS the key applies nothing from it, and still reports as a non-fork. This is the misreading the key was renamed to prevent, so testing only the positive would have been the easy miss.
- **"Declaring the entry is CHEAP."** This is the one that forced a real type change rather than an addition: `chain` was required, so an entry existing only to carry `whenForked` would not have compiled. It is now optional, which widens the input and leaves every existing configuration valid, and the changeset calls it out. Tested with an entry that carries the fork layer and nothing else.
- **"The local endpoint no longer comes from the local chain bucket"**, with the zero-configuration path unchanged. Tested both ways: a fork ignores a `chains[31337]` endpoint and dials the conventional local one, including with no chain configuration whatsoever, while a NON-fork run on that same local chain still dials its own configured endpoint, so the change is scoped to forks.

The layering (`chains[<id>]`, then `overrides`, then `whenForked`) is applied consistently to both the connection config and the semantics, and is tested in order.

**Off-path finding, correctly captured rather than fixed:** the note added in `work/notes/observations/` records that the `overrides` layer still supplies a fork's endpoint, so on a real network's environment that endpoint is the real network's. It is verified empirically, correctly scoped as latent (the only caller that can fork always passes a provider, which wins the merge) and correctly left out of scope here. It becomes reachable on the future `--is-fork` path, so it is worth tasking.

PR #116, opened 2026-08-28
feat(document-fork-runs): Document what a fork run is, what it inherits…

wighawag:


wighawag:
APPROVE ✅ (conductor Gate-3: page reviewed against the task's acceptance criteria)

**Verification.** The runner's gate bounced only on the known `@rocketh/export` in-process compile timeouts, so I ran the configured chain myself on the rebased tip in a clean scratch clone:

- `pnpm test`: 1098/1102, the only failures being those export timeouts
- `packages/rocketh-export` alone: it flaked again under machine load, so I re-ran it with `--testTimeout=60000` and got **35/35**, which isolates the 5s budget as the sole cause

**The criteria most at risk of passing on paper, for a docs task, are accuracy ones:**

- **"Every configuration sample matches the shipped types rather than being written from memory."** The samples use `whenForked` (not the `fork` spelling this task was written with, which the shipped type does not have), keep `chain` where it is load-bearing, and follow the repo's `as const satisfies UserConfig` form.
- **"Inheritance is documented as CONDITIONAL, with the declaration that fixes the hardhat case."** The three-source order is stated in the right order, anvil is correctly described as needing nothing, and the hardhat case is named as landing on the dev-node bucket with the one-line remedy shown. This is the criterion a page could most easily satisfy by promising inheritance unconditionally, and it does not.
- **"The saving non-goal is stated precisely."** It says the right thing: core would save, the hardhat plugin suppresses it in the caller, so anyone driving core directly must pass `saveDeployments: false` themselves. Understating this would have been worse than omitting it.

The two-chain-ids section is framed as two questions with two answers, and it draws a consequence I had not seen stated anywhere: an `accounts`/`data` entry keyed by chain NUMBER follows the node, so `{1: ...}` resolves on an anvil fork and falls through to `default` on a hardhat one, with the remedy of keying by environment name. That is the kind of thing a reader hits and cannot diagnose. No em dashes, paragraphs not hard-wrapped, and the unknown-signers page gained pointers in both directions.

**Off-path finding, correctly captured rather than fixed:** the note recording that the existing hardhat-deploy fork-testing guide destructures `env.network.name`, which does not exist on `Environment['network']`, so that snippet logs `undefined`. Worth tasking as a docs fix.

PR #120, opened 2026-08-29
feat(a-fork-does-not-save-unless-asked): Move "a fork does not save" in…

wighawag:
Task: `work/tasks/done/a-fork-does-not-save-unless-asked.md`

All checks green: `pnpm typecheck`, `pnpm build`, `pnpm format:check`, `pnpm test` (97 files, 1123 tests).

## What I did

**Drift check first (all three premises confirmed against current code):** `saveDeployments` still defaulted in `resolveExecutionParams` (`packages/rocketh/src/executor/index.ts`) as described; the no-provider branch still short-circuited to `true` before the `memory`/`hardhat`/`default` name check; `packages/hardhat-deploy/src/helpers.ts:130` still passed `saveDeployments: isFork ? false : undefined`.

**Core (`packages/rocketh/src/executor/index.ts`).** Added a fork term to the `saveDeployments` default, placed above BOTH existing branches: `if (fork) false; else if (!provider) true; else <existing name check>`. The explicit `executionParameters.saveDeployments` is still read first, so an explicit `true` keeps saving on a fork. No new destination for fork records was invented.

**Test-first at the seam.** `packages/rocketh/test/fork-does-not-save.test.ts` (13 tests) was written before the caller changes and verified RED two ways: with the fork term removed (4 failures), and with the fork term placed only inside the named-environment branch (2 failures, both on the no-provider path) - so the "fix one branch only" mistake is caught. It covers: fork/no-fork contrast with a provider; the same contrast with NO provider; a fork input with no chain id; explicit `true` and `false` on both branches; the three ephemeral names, a named environment, and the no-provider short-circuit for non-fork runs; and, through a real `loadEnvironmentFromStore`, that a fork rehearsal's `env.save()` writes nothing to the store while `env.deployments['MyContract']` still loads the forked network's record on a node reporting 31337 (the read path, untouched).

**Callers.** `helpers.ts` no longer passes `saveDeployments` at all. `tasks/deploy.ts` keeps its `isFork ? false : saveDeployments` clamp (see Decisions). `packages/hardhat-deploy/test/fork-does-not-save.test.ts` (4 tests) mocks `@rocketh/node` to capture what the real `loadEnvironmentFromHardhat` passes (asserting `saveDeployments` is ABSENT, not `false`), then feeds those exact params to real core: fork resolves to no saving and still reads mainnet's records; a plain run still saves.

**Docs / conventions.** `documentation/fork-runs/index.md`: the "Saving is unchanged" section is replaced by "A fork does not SAVE, and you do not have to remember that", with the deliberate-save escape hatch shown, and the now-unnecessary `saveDeployments: false` lines removed from two examples; the parent heading became "One thing a fork run does not do yet". ADR 0014 gained a "Refinement" section (the file's existing practice for closing a deferred hazard) and its Consequences bullet now points at it. `CONTEXT.md`'s **fork run** glossary entry gained a one-clause "records flow ONE WAY" note. Changeset added at `.changeset/a-fork-does-not-save-unless-asked.md` (`rocketh`: minor, `hardhat-deploy`: patch).

Working tree contains only the intended changes; no scratch files.

## Decisions

- **Where the fork term sits: ABOVE both existing default branches, not inside either.** The resolution is now `explicit param` → `fork ? false` → `no provider ? true` → `name check`. I considered adding it to the named-environment branch (the branch the trap is usually described in), but the no-provider branch short-circuits to `true` before the name is ever consulted, and a fork driven WITHOUT a provider is exactly the `--is-fork` shape (attach to an anvil fork by rpc url), so that placement would have left the hazard live on the one path the rule exists to protect. It is pinned by a test that fails under that alternative. _Touches:_ the future `--is-fork` CLI task, which can now pass a fork input with no provider and inherit the safe default with no extra flag handling.
- **The `helpers.ts` caller guard is REMOVED; the `tasks/deploy.ts` one STAYS.** They are not the same shape, which the task's framing (one guard) does not capture. `loadEnvironmentFromHardhat` passed `isFork ? false : undefined`, so deleting it changes nothing now that core answers `false` for a fork, and deleting it is what proves the rule really moved; passing `false` "as a belt" would hide a future core regression behind a local guard, so leaving it was rejected. `tasks/deploy.ts` is different: its `saveDeployments` is an explicit boolean by construction (it starts at `true`), and an explicit value outranks every default, so core's rule cannot reach that call site at all and removing the clamp would turn a `HARDHAT_FORK` run into an explicit `true` and write into the forked network's records. Both call sites carry a comment saying which case they are. _Alternative considered and rejected:_ reworking `tasks/deploy.ts` to leave the value `undefined` when `--save-deployments` is absent, so core decides there too. It would genuinely centralise the rule, but it also changes NON-fork behaviour (a hardhat network literally named `memory`/`hardhat`/`default` on a non-EDR connection would flip from saving to not saving), and the acceptance criterion is that the hardhat-deploy path behaves exactly as today. _Touches:_ the `--save-deployments` flag, which stays inert on a fork exactly as today; making that flag the fork escape hatch would be a user-visible change to the flag itself and belongs with the `--is-fork` work, not here.
- **ADR 0014 was amended rather than left alone**, with a "Refinement" section plus a "(Closed since…)" clause on the deferral bullet, following the precedent set by commit `54233e9a`, which closed the `overrides.rpcUrl` hazard the same way in the same file. The alternative was a new ADR; this is a consequence of ADR 0014's existing model rather than a new decision, and splitting it would scatter the fork model across two records.


wighawag:
APPROVE ✅ (conductor Gate-3: diff reviewed against the task's acceptance criteria; the acceptance gate ran green first time, no re-verification needed)

- **"This holds on the no-provider path as well."** The fork term is placed ABOVE both existing branches, not inside the named-environment one, which is the trap: the default short-circuits to `true` when there is no provider, and a fork with no provider is exactly the `--is-fork` case this rule exists to protect. Tested on both paths, so a fix to one would not pass.
- **"A fork still READS the forked network's records."** Kept as a live test rather than an assumption, which matters because the read path sits next to the code being changed and is the property the whole feature exists for. The payoff test asserts both halves at once: a save during a rehearsal stays in memory while the on-disk records are untouched, and the run still reads them.
- **"The hardhat-deploy path behaves exactly as it does today."** This is where the task was incomplete and the work was better than it: there are **two** call sites, not the one I named. `helpers.ts` passed `undefined` off a fork, so its guard is correctly deleted and core's rule now covers it. `tasks/deploy.ts` is correctly left alone, and the reason given is right, which I verified rather than took on trust: `saveDeployments` starts at `true` on line 23 and is never `undefined` at the call site, so an explicit value always outranks core's default there and deleting the guard would have sent `true` and written into the forked network's records. Getting that backwards would have been a silent regression with no test failing.

Also correct: the explicit override still wins in the reachable direction, non-fork behaviour is pinned including the ephemeral names, and no fork-saves-elsewhere destination was invented.

The documentation was updated rather than left to contradict the code: both samples drop the now-redundant `saveDeployments: false`, the "Saving is unchanged" non-goal becomes a positive section with the escape hatch, and "Two things a fork run does not do yet" correctly becomes one. ADR 0014's consequence line is annotated rather than rewritten, so the original decision stays visible, and CONTEXT.md gains the one-way records rule.

PR #121, opened 2026-08-29
feat(a-provider-less-fork-discovers-its-connected-chain-id): A fork run…

wighawag:
Task: `work/tasks/done/a-provider-less-fork-discovers-its-connected-chain-id.md`

Done. Everything is green: `pnpm typecheck`, `pnpm build`, `pnpm format:check`, `pnpm test` (98 files / 1138 tests).

## What I checked first (drift)

- `getChainIdForEnvironment` (`packages/rocketh/src/executor/index.ts`) did still dial only when `executionParams.provider` was supplied (`const chainIdFromProvider = provider ? ... : undefined`), and with no provider its only source was `environments[<name>].chain`, throwing `Could not find chainId ... (no provider)`. Premise confirmed.
- A fork's endpoint in `resolveExecutionParams` was assembled from three places: `getChainConfigFromUserConfig(..., fork ? CONVENTIONAL_LOCAL_RPC_URL : undefined)` (so `chains[<id>].rpcUrl` is bypassed), `overrides.rpcUrl` withheld on a fork, then the `whenForked` layer spread on top. Net answer `whenForked.rpcUrl` else `CONVENTIONAL_LOCAL_RPC_URL`, as the task said. `ChainUserConfig` has no `provider` field, so those layers cannot supply a connection any other way.
- Note: the task body lives at `work/tasks/backlog/a-provider-less-fork-discovers-its-connected-chain-id.md`, not `ready/`. I left it where it is and did no git.

## What I changed

`packages/rocketh/src/executor/index.ts`

- New `resolveForkRunEndpoint(config, forkedNetworkName)`: the single answer to "where does a fork run connect", `environments[<name>].whenForked.rpcUrl` else `CONVENTIONAL_LOCAL_RPC_URL`. Both `resolveExecutionParams` (as `rpcUrlInsteadOfConfigured`) and the new discovery call it, so they cannot drift.
- New private `askForkNodeForItsChainId(endpoint, environmentName)`: builds a `JSONRPCHTTPProvider` on that endpoint and asks `eth_chainId`; throws naming the endpoint when the node is unreachable or answers something that is not a chain id.
- `getChainIdForEnvironment` now reads `chainIdFromNode = provider ? <ask provider> : fork ? <ask the fork node> : undefined`. The guard is fork-only and the comment states it as a necessity (off a fork the endpoint comes from `chains[<id>]`, so asking first is circular). Adoption order is unchanged, so the node's answer still wins over the declared id; the identity-check leniency (`!fork`) and the final error are untouched. `resolveForkDescriptor` was not touched, so the SIMULATED identity is exactly as before.

`packages/rocketh/test/fork-connected-chain-discovery.test.ts` (new, written before the production change and confirmed red: 11 failing / 4 passing, then all 15 green). It stubs only `fetch`, so the real `JSONRPCHTTPProvider` runs and every "which endpoint" assertion is made on the url production code asked for. It covers: discovery with nothing declared; zero-config end to end (node reports 1, `chains[1]` settings found with no `environments` entry, local endpoint, `chain.id === 1`); a `whenForked`-only entry; the node winning over a declared `chain: 1` (the hardhat 31337 shape) with the descriptor still `{networkName: 'mainnet', chainId: 1}`; a supplied fork `chainId` still naming the simulated chain; discovery dialling the exact endpoint the run then connects to on port 8546; never dialling `overrides.rpcUrl`; the non-fork provider-less run still throwing the existing error and dialling nothing; unreachable/nonsense answers failing with the endpoint named; and provider-supplied runs dialling nothing.

`documentation/fork-runs/index.md`: the "One thing to declare on this path" paragraph documented the exact behaviour this replaces (it told users the run fails when nothing declares a chain), so it would have been actively wrong. It now says the run asks the node, that the node's answer wins over `chain`, and that an unreachable node stops the run naming the endpoint. The `whenForked`-only claim further down was left alone and is now pinned by a test, per the task.

`.changeset/fork-run-discovers-its-connected-chain-id.md`: `rocketh` minor.

`work/notes/observations/fork-docs-still-say-overrides-rpcurl-is-dialled.md`: out-of-scope drift I noticed while editing that page (the docs still tell users to restate the fork endpoint because `overrides.rpcUrl` "would otherwise be what a fork run dials", which the endpoint refinement already withholds). Captured, not fixed.

## Decisions

- **An unreachable node FAILS the run; it does not fall back to the declared id.** The only fallback available is `environments[<name>].chain`, and on a fork that value routinely names the _simulated_ network (the docs tell a hardhat user to declare `chain: 1` so mainnet's settings are found), so falling back would silently hand the _connected_ identity a simulated id and every transaction would be signed for a chain the node rejects. That is exactly the bug `fork-chain-identity-simulated-versus-connected` closed. A fork run that cannot reach its node can do nothing useful anyway, so stopping loses nothing. The alternative considered was falling back to the declared id with a warning, rejected for the reason above. Both error messages name the endpoint that was tried and the message for an unreachable node points at `environments.<name>.whenForked.rpcUrl`. This introduces a new refusal on a path that previously threw anyway (it threw `Could not find chainId ... (no provider)` before dialling), and it touches the future `--is-fork` flag: `rocketh -e mainnet --is-fork` against a node that is not up will fail with the endpoint named rather than proceeding on a declared id.
- **A node answer that is not a usable chain id is treated as the same failure.** `0`, `NaN` or a non-integer throws naming the endpoint, rather than falling through the `||` to the declared id, which would reopen the same wrong-identity path by a side door. Alternative considered: leave the existing `||` semantics and let a nonsense answer fall back. Rejected as the cheaper-looking but more dangerous branch.
- **The shared endpoint resolution lives in `packages/rocketh/src/executor/index.ts` as `resolveForkRunEndpoint`, next to `CONVENTIONAL_LOCAL_RPC_URL`,** which I moved up in the file so both consumers read after its definition. That module already owns both callers (`getChainIdForEnvironment` and `resolveExecutionParams`) and already owned the constant, so no new module or dependency edge was needed. The alternative considered was `packages/rocketh/src/environment/chains.ts` alongside `getChainConfigFromUserConfig`; rejected because that module is keyed by chain id and knows nothing about environments or fork runs, so the fork-only rule would have sat at the wrong layer. It is exported from the executor module (like its neighbours) but deliberately NOT re-exported from the package root `packages/rocketh/src/index.ts`: nothing outside needs it yet, and the resolved endpoint is already observable on `resolvedExecutionParams.provider`.
- **Naming: `resolveForkRunEndpoint`, not `resolveForkRpcUrl`.** `CONTEXT.md` and ADR 0014 pin that the bare word _fork_ is ambiguous across three parts of speech and that a `FORK_RPC_URL` reads as the url to fork FROM, which is the one thing this is not. _fork run_ is the existing glossary term for the adjective reading, so the name reuses it rather than coining a new concept, and the JSDoc says so at the choice site.


wighawag:

- **"The node's answer WINS over `environments[<name>].chain`."** This is the crux and the easy thing to get backwards, because preferring the declared value looks like a free optimisation. It would have reintroduced exactly the bug `fork-chain-identity-simulated-versus-connected` closed, since the docs tell a hardhat user to declare `chain: 1` and that id would then have been signed into transactions sent to a node reporting 31337. The order is right and the reasoning is stated at the site. Tested with a declared id that differs from the node's, and paired with a test that the SIMULATED id still comes from the declared value, so the two identities are pinned apart rather than merely one of them pinned.
- **"Discovery dials the SAME endpoint the run then connects to."** Closed structurally rather than by convention: `resolveForkRunEndpoint` is one function and both `getChainIdForEnvironment` and `resolveExecutionParams` now call it, so they cannot drift. Tested on a non-conventional port, plus the sharper negative that it never dials the simulated network's endpoint.
- **"Discovery is scoped to forks."** The guard reads as the necessity it is (off a fork the endpoint lookup needs the chain id, so the question is circular), and the provider-less non-fork path is tested to still throw *without dialling*, which is what proves the scoping rather than just the outcome.

The unreachable-node decision is the right one and well argued: falling back to the declared id was the dangerous option precisely because that value routinely names the simulated network, so stopping is honest. The error names the endpoint and points at `whenForked.rpcUrl`, and a node answering something that is not a chain id is rejected rather than coerced.

The documentation section that told users to declare `chain` on this path is removed rather than left to contradict the code, and it correctly says the node's answer wins.

**Off-path finding, correctly captured rather than fixed:** the note that the `whenForked` section still tells readers to restate the endpoint because `overrides.rpcUrl` would otherwise be dialled. That prose went stale when #119 withheld it, so this is my staleness, not this task's, and it was right to capture rather than widen scope. I will fix it directly.

PR #122, opened 2026-08-29
feat(is-fork-flag-on-the-cli): Give rocketh users who are not on hardha…

wighawag:
Task: `work/tasks/done/is-fork-flag-on-the-cli.md`

All acceptance criteria are met and the full gate is green (`format:check`, `typecheck`, `build`, `test` 1152/1152, `sync:template:check`, `docs:build`, `test:getting-started`).

## What I did

**Drift check first.** Both blockers are landed in core on this branch: the saving rule (`resolveExecutionParams`, above both default branches, `packages/rocketh/src/executor/index.ts`) and provider-less fork chain-id discovery (`askForkNodeForItsChainId` + `resolveForkRunEndpoint`). The CLI still builds its params by spreading `options`. No drift; the earlier refusal's false premise is genuinely fixed. (Note: the prior attempt's branch carries no commits over `origin/main`, so there was no prior code to build on, only its recorded refusal. Also, the task body currently sits in `work/tasks/backlog/`, not `ready/`; I did not move it.)

**Test first** (`packages/rocketh-node/test/cli-is-fork-flag.test.ts`, 14 tests, written red then green, and checked to be discriminating by reverting the mapping: 9 of them fail).

**Production changes**
- `packages/rocketh-node/src/cli-options.ts` (new): `buildCLIProgram(version)` (the whole option surface, now parseable by a test) and `resolveEnvironmentInput(options)` mapping `-e <name> --is-fork` to `{fork: <name>}`.
- `packages/rocketh-node/src/cli.ts`: uses the builder, and sets `environment: resolveEnvironmentInput(options)` **after** the spread, next to the `onUnknownSigner` fixup and commented for the same reason.
- `documentation/fork-runs/index.md`: new "From the command line: `--is-fork`" section as the first of three ways to invoke a fork, plus the `--save-deployments` escape hatch from the CLI; both places that said the flag does not exist are gone.
- `packages/rocketh-node/README.md`: option-table row + a short "Rehearsing on a fork" section.
- `.changeset/is-fork-flag-on-the-cli.md`: `@rocketh/node` minor.

**Live confirmation** (the criterion said confirm end to end, not assume): the built `dist/cli.js` was run against a real anvil from a throwaway project outside the repo, in the three shapes (fork on another port via `whenForked.rpcUrl`; the same command without the flag; zero configuration on the conventional endpoint). Transcript kept at `docs/spikes/is-fork-flag-on-the-cli/live-anvil-smoke.md` and referenced from the test file. Port 8545 was already held by an unrelated project's `hardhat node`, which I left alone and used read-only; my own anvil on 8546 was stopped and the scratch project deleted.

**Out-of-scope signal captured**: `work/notes/observations/cli-tags-option-is-not-split.md` (the CLI's `--tags` reaches core as a string, so the tag filter iterates characters; hardhat-deploy splits, the CLI does not). Not fixed.

## Decisions

- **`environment` is transformed AFTER the spread, and the ordering is pinned by a test rather than by a comment alone.** The CLI does `...(options as ExecutionParams)` and then fixes up individual fields; commander's `environment` is _always_ a string, so a fork input built before the spread is silently overwritten and the run is quietly not a fork. I put the transform next to the existing `onUnknownSigner` fixup, with the same reasoning in the comment, and added a test that asserts `{...options}.environment` is the raw string while the built params carry the `ForkInput`. The alternative (normalise the options bag before spreading) would have made the two fixups live in different places for no gain. _Touches_: nothing else; `--save-deployments`, `--reset` and `--on-unknown-signer` keep their current handling.
- **The CLI's option surface moved into `packages/rocketh-node/src/cli-options.ts`, verbatim, so the flags can be parsed by a test.** `cli.ts` is a bin script that executes a deployment on import, so its flags were reachable by no test at all: "the flag takes no argument", "`--fork` is not an alias" and the `--help` wording are acceptance criteria that can only be checked by parsing argv. `cli.ts` still owns everything with side effects (log level, `resolveOnUnknownSigner`, the executor call). The alternative was spawning the built CLI as a subprocess per case, which is slower, needs a project on disk, and still cannot inspect the option list. _Touches_: every other CLI flag moved with it (no behaviour change, verified by `--help` output and the live runs); `resolveOnUnknownSigner` deliberately did NOT move, since it calls `process.exit` and is another task's surface.
- **No `chainId` is inferred for the fork input, deliberately.** `@rocketh/node` does hold a name-to-chain map, but it is keyed by viem's kebab-cased chain name (mainnet's key is `ethereum`), so matching it against an environment name would be a coincidence dressed as a mechanism (ADR 0014 says so explicitly). The CLI therefore emits the bare `{fork: <name>}` and lets the run ask the node. _Touches_: the chain-id discovery path landed by `a-provider-less-fork-discovers-its-connected-chain-id`, which is what makes the zero-configuration case work; the hardhat-node caveat (declare `environments[<name>].chain`) stays the documented remedy.
- **`--save-deployments` stays the only way to turn saving on for a fork; I did not add `--no-save-deployments`.** The flag is set-only, and once the fork default is off that is exactly the right shape (ADR 0014's saving refinement says as much). Documented on both the docs page and the README instead of adding a flag. _Touches_: a hypothetical future negative flag, and the `--fork` name, which stays reserved.
- **The `@rocketh/node` README was updated too, although the criterion named only `documentation/fork-runs/index.md`.** That README carries the CLI's option table, so shipping a flag absent from it would leave a second page quietly denying the feature, which is the exact failure the criterion exists to prevent. _Touches_: the `@rocketh/node` package docs; covered by the same changeset.


wighawag:

- **"Confirm the zero-configuration path end to end rather than assuming it."** This is the one a task like this usually satisfies with a unit test and a hopeful sentence, and it did not: `docs/spikes/is-fork-flag-on-the-cli/live-anvil-smoke.md` records the BUILT bin script driven against real nodes from a throwaway project, on both fork shapes. The anvil shape adopts `1`, inherits mainnet's tags rather than the dev node's `local`, reads `deployments/mainnet` and writes nothing back; the hardhat shape adopts `31337` with nothing declared at all. It also ran the control, the same command WITHOUT the flag, which fails with the pre-existing no-provider error and is the sharpest available proof that the flag, not the configuration, is what makes a run a fork.
- **"The flag is spelled `--is-fork`; `--fork` is not introduced as a name or an alias."** Tested as a negative rather than asserted, and the `--help` text is pinned to the assertion voice, which is the part that would rot silently.
- **"Transform `environment` relative to the spread."** The failure mode here is invisible (the spread would overwrite a fork input built before it, leaving a run that silently is not a fork), so it is pinned by its own regression test rather than left to the comment.

The option surface was extracted into `cli-options.ts` so the flags can be parsed by a test at all, which is the right seam: `cli.ts` is a bin script that runs a deployment on import, so everything it declared was previously pinned by nothing. Correctly, no `chainId` is inferred from the environment name, and the reason given is the real one: the name-to-chain map is keyed by viem's kebab-cased name, so mainnet's key is `ethereum` and any match would be a coincidence. The docs now document the flag in both places that previously denied it existed, including that `--save-deployments` is the CLI escape hatch.

**Off-path findings, correctly captured rather than fixed.** One is worth acting on soon: `cli-tags-option-is-not-split` reports that the CLI hands `--tags Token` to core as a STRING while `tags` is `string[]`, so the tag filter iterates characters and no script tag can ever match. If that reproduces it means `--tags` has never worked on this CLI, while hardhat-deploy splits correctly. It was right to keep it out of this task, and it should be tasked.

PR #124, opened 2026-08-29
feat(cli-tags-reaches-core-as-a-list): The CLI `--tags` value never rea…

wighawag:
Task: `work/tasks/done/cli-tags-reaches-core-as-a-list.md`

Done. Everything green: `pnpm typecheck`, `pnpm build`, `pnpm format:check`, `pnpm test` (100 files / 1169 tests).

## Drift check (done first, against current code, not the snapshot)

- `packages/rocketh-node/src/cli.ts` did still spread `...(options as ExecutionParams)` with fix-ups only for `onUnknownSigner` and `environment`; no `tags`.
- `packages/rocketh/src/executor/index.ts:684` did still do `for (const tagToFind of resolvedExecutionParams.tags) { ... if (tag === tagToFind) }`.
- I re-ran the reproduction rather than trusting it: with the pre-fix pass-through in place, my new end-to-end tests confirm `--tags Token` runs **nothing** and `--tags cat` runs the script tagged `a`. Note `--tags a,b` **passes** even with the broken code (chars `a`, `,`, `b`), exactly as the task predicted, so the single-character case is the discriminating one.
- History check for the changeset: the split was NOT always absent. `readConfig` split on `,` from `7cb42855` (2023-06-06) and was removed in `e2dbd6f7` ("revamp of types and resolution", 2025-09-19). This is a **regression**, and the changeset says so.

## What I changed

- `packages/rocketh-node/src/cli-options.ts` — added `RockethCLIOptions` (the parsed shape), `parseTags`, `resolveUnknownSignerPolicy`, and `toExecutionParams` (the whole boundary, one explicit entry per option); `resolveEnvironmentInput` lost its `as string`.
- `packages/rocketh-node/src/cli.ts` — the blanket cast and the post-spread fix-ups are gone; it parses, sets up the logger, calls `toExecutionParams`, and turns its refusal into the same stderr message + exit 1 (verified against the built `dist/cli.js`).
- `packages/rocketh-node/test/cli-tags.test.ts` (new) — written before the fix. Selection cases drive real script files through argv → commander → mapping → `loadAndExecuteDeploymentsFromFiles` → the real selection loop.
- `packages/rocketh-node/test/cli-is-fork-flag.test.ts` — its local copy of the CLI's spread now calls the production mapping.
- `.changeset/cli-tags-reaches-core-as-a-list.md`, and `work/notes/observations/execution-params-network-alias-cast.md`.

## Other mismatch the mapping revealed (named per the acceptance criteria)

`--scripts` and `--deployments` were also silently dropped — same commit, same cast, different reason: they are `ConfigOverrides` in core, not run parameters, so the spread left them as excess properties `resolveExecutionParams` never reads. Both documented flags did nothing. Fixed (see Decisions). Nothing else surfaced; `logLevel` has no core home by design and is consumed by the bin script.

hardhat-deploy is untouched.

## Decisions

**Fixed at the CLI boundary, not in core.** `parseTags` splits in `@rocketh/node`, leaving `ExecutionParams.tags: string[]` honest. Alternative considered: coercing `string | string[]` in `resolveExecutionParams`, which would protect every caller but makes the declared `string[]` a lie and invites the next caller to pass a string too. It also puts a CLI-syntax concern (comma separation) in core, which knows nothing about command lines and already _refuses_ a tag containing a comma. Touches: nothing else, since core's contract is unchanged.

**The typed mapping is `toExecutionParams(options: RockethCLIOptions): ExecutionParams`, in `cli-options.ts`.** Two named types plus one total function, rather than per-field fix-ups after a spread. `RockethCLIOptions` declares what commander actually parses (`tags?: string`, flags `true`-or-absent, `environment` required) and `program.opts<RockethCLIOptions>()` is the single remaining assertion, made once and stated, instead of once per run and hidden. Because every field is written out, a CLI-shape-vs-core-shape mismatch is now a compile error. Two enforcements back it: a `@ts-expect-error` pinning that `tags: 'a,b'` must _not_ typecheck (`pnpm typecheck` covers `test/`, per `CONTEXT.md`), and a `Record<keyof RockethCLIOptions, true>` compared against the program's declared options, so declaring a flag without declaring its parsed shape goes red. Touches: `cli-is-fork-flag.test.ts`, which mirrored the old spread and now calls the mapping.

**`--scripts` / `--deployments` fixed rather than left as an observation.** The task's rule was "fix if trivially the same shape". It is: they belong in `ExecutionParams.config` (`ConfigOverrides`), and the sibling CLIs (`@rocketh/export`, `@rocketh/doc`, `@rocketh/verifier`) already route `--deployments` there, so the destination was not a judgement call. Alternative considered: encoding the drop explicitly in the mapping and filing an observation, rejected because it would have written the bug down as intent while the correct target sat one field away. Note this is user-visible: `rocketh -d somewhere` was ignored and now moves where deployments are read and written. `config` is always present with possibly-`undefined` entries; `resolveConfig` skips undefined, so an absent flag still overrides nothing. Touches: any caller relying on those flags being inert (none in this repo), and a pre-existing precedence subtlety I did _not_ change, that `environments[<name>].scripts` still wins over `config.scripts` inside `resolveExecutionParams`.

**Comma segments are not trimmed.** `--tags "a, b"` yields `['a', ' b']` and ` b` matches nothing. I matched hardhat-deploy exactly (guard `''` first, then `split(',')`) rather than adding a trim, because the task named that implementation as the intended shape and because silently editing a tag is another way for the flag to mean something other than what the user typed. Alternative considered: trim and drop empty segments, friendlier but a divergence from the reference behaviour and a second place where the flag's meaning is decided. Recorded because it is a user-visible choice, and it is cheap to reverse if someone prefers the friendlier reading.

**The `--on-unknown-signer` refusal moved from `process.exit` to a `throw`, caught in `cli.ts`.** Same message on stderr, same exit code 1 (checked against the built `dist/cli.js`), but the option surface is now callable from a test, which is what let the acceptance criterion "the already-correct fix-ups keep working, tested" be met. This is a relocation of an existing refusal, not a new one. Touches: `cli.ts` only, which is the one place allowed to talk to a terminal.


wighawag:

- **The single-character case.** This is the one that separates a real fix from a plausible one, since splitting alone passes a naive test while a character-iterating implementation still fails here. It is tested, and so is `--tags ''` meaning no filter rather than a filter for the empty tag, which is the trap a bare `.split(',')` walks into.
- **"State honestly whether this was a regression or never worked."** It checked rather than assumed, and the answer contradicts what I assumed when writing the task: it is a REGRESSION. I verified the attribution independently — `e2dbd6f` (2025-09-19, "revamp of types and resolution") removed the line `tags: ... options.tags.split(',')`, and the cast is what let the loss type-check. My "probably never worked" framing was wrong and the changeset is right.
- **"The class is closed, not just the instance."** The typed mapping alone would have been a convention. It is backed by two enforcements pulling in opposite directions: a `@ts-expect-error` pinning that `tags: 'a,b'` must not typecheck, and a `Record<keyof RockethCLIOptions, true>` reconciled against the program's declared options, so declaring a flag without declaring its parsed shape goes red. I checked the `@ts-expect-error` is not inert: `packages/rocketh-node/tsconfig.json` includes `test/**/*.ts`, so `tsc --noEmit` covers it and an unused directive is itself an error.

**The finding that makes this bigger than the task.** Removing the cast revealed that `--scripts` and `--deployments` were silently dropped too, by the same mechanism for a different reason: they are `ConfigOverrides` in core, so the spread delivered them as excess properties nothing reads. I confirmed independently that nothing ever read them at top level and that the old CLI never set `executionParams.config`, so `rocketh -d somewhere` really was inert. Fixing them here was the right call under the task's "trivially the same shape" rule, since the sibling CLIs already route `--deployments` to the same place, and the report correctly flags it as user-visible: that flag now moves where deployments are read and written.

Also good: the non-trimming of comma segments is a deliberate, recorded choice matching hardhat-deploy rather than a silent divergence, and moving the `--on-unknown-signer` refusal from `process.exit` to a caught `throw` is what made "the existing fix-ups keep working" testable at all, with the message and exit code checked against the built `dist/cli.js` rather than assumed.

**Off-path finding, correctly captured rather than fixed:** `ExecutionParams` still carries an undocumented `network` alias read through `(executionParams as any).network`. That is the same class of hole this task just closed, one layer down, and it is worth a decision (declare it deprecated, or drop it).

PR #129, opened 2026-08-30
feat(capture-broadcast-transactions-on-the-environment): Capture every…

wighawag:
Task: `work/tasks/done/capture-broadcast-transactions-on-the-environment.md`

All checks are green. Here is what I did.

## What I built

**`packages/rocketh-core/src/types.ts`**

- New exported entry type `CapturedTransaction`: `{from} & ({type: 'intent'; to?; value?; data?; signability} | {type: 'raw'; raw})`, documented with why there is no account name, no gas/fee/nonce/hash/receipt, why `value` stays a 0x quantity, and why `signability` is on the intent arm only.
- New field on `Environment`: `readonly capturedTransactions: readonly CapturedTransaction[]`, alongside `deployments` and `tags`. Nothing else in the interface was reshaped.

**`packages/rocketh/src/environment/index.ts`**

- A module-private `toCapturedIntent(transactionData, signability)` builder (functional, no class) that reads the transaction as the choke point _received_ it, keeps `from` verbatim, and leaves absent fields absent.
- A per-run `const capturedTransactions: CapturedTransaction[] = []`, exposed by identity on the environment.
- Four capture sites, all inside the private `broadcastTransaction` choke point, each on the success path of its own arm: the raw relay arm, the interactive (`ask`) resolution, the `wallet`/`remote` arm and the `signerOnly` arm. I confirmed by grep that these three send sites (`eth_sendRawTransaction` x2, `eth_sendTransaction`) are the only ones in any package's `src`, so no transaction reaches a node without passing capture.

**`packages/rocketh/test/captured-transactions.test.ts`** (new, written before the production code; 18 tests, all red first, then green): order across a mixed run through both public funnels, the in-process list off the environment the executor returns, concurrent broadcasts recorded in observed-completion order, exact key sets per arm, `to` absent for a creation, `data` absent for the funding transfer, `value` as a 0x quantity plus a `JSON.stringify` round trip, intent-not-signed-payload for a locally signed send, `from` kept checksummed while signability still resolves, raw entry carrying `from`/`raw` only (with the relayer's would-be `unsignable` pinned as the reason), no entry for a deferred `throw`, exactly one `unsignable` entry for an `ask`-resolved transaction, no entry for a failed send (object and raw), capture on both a memory run and a fork run, and an impersonated/node-held split a batch consumer can read.

**`.changeset/captured-transactions-on-the-environment.md`**: `minor` for `@rocketh/core` and `rocketh`.

Gates: `pnpm build`, `pnpm typecheck`, `pnpm test` (101 files, 1192 tests) and `pnpm format:check` all pass. `pnpm sync:template:check` passes. `pnpm changeset status --since=main` currently reports "no changesets found" only because the new changeset file is still untracked (that command diffs git, which does not see untracked files); it resolves once the runner commits with `git add -A`. I did not run `pnpm test:getting-started` (it scaffolds a project over the network and is untouched by this change).

## Decisions

1. **The field is `env.capturedTransactions`, and the entry type is `CapturedTransaction`.** Chosen for coherence with the language the spec and the two follow-on tasks already use (_captured transactions_, and `document-captured-transactions` asks for a `CONTEXT.md` glossary entry for _a captured transaction_ distinguishing it from the pending-transaction record). Alternatives considered and rejected: `transactions` (too generic beside `network.provider.transactionHashes`, and it says nothing about the run-scoped, what-we-DID meaning), `broadcastTransactions` (reads as a verb, the same ambiguity `CONTEXT.md` pins against for `--fork` vs `--is-fork`), and `broadcasts` (loses the noun the consumers actually want). It touches the file-sink task (which serialises this list) and the documentation task (which pins the term), so both inherit this spelling.

2. **`data` is OPTIONAL on the intent arm.** The shape block quoted in the task types it as required (`data: 0x${string}`), while the same task's body requires the deterministic-factory funding transfer, which genuinely carries none, to have _no `data` key at all_. `EIP1193TransactionData.data` is itself optional (`eip-1193@0.6.5`, `BaseEIP1193TransactionData`), so a required field would have been a lie the compiler could not keep. I followed the body and the acceptance criterion (absent stays absent) over the literal type block. This touches the file-sink task: a serialiser must handle an intent entry with no `data`, and must not substitute `'0x'` (which a replay would send as an empty call rather than a plain transfer).

3. **Capture happens immediately after the send RPC resolves, BEFORE the `autoMine` `evm_mine` call.** The truth being recorded is "the run sent this", and the node accepted it at that point; if a subsequent `evm_mine` fails the run throws with the entry already present, which is correct, because the transaction really was sent. The alternative (capture after `autoMine`, just before each `return`) would drop an entry for a transaction that was genuinely broadcast. This also fixes the concurrency promise concretely: the order is the order the run observed the _sends complete_.

4. **On the `ask` path the entry is pushed at one single site, after the pasted transaction has been found, checked successful and (for a deployment) checked to have deployed something, just before the hash is returned.** Every earlier exit from that branch either defers (`throw` policy, "cannot sign", a paste the user would not vouch for) or fails verification, so none of them can leak an entry, and the successful path cannot produce two. What is recorded is the intent _rocketh asked for_, not the transaction the human actually sent, since a Safe execution goes to the Safe carrying the call inside it and is not replayable as such.

5. **The array is exposed by IDENTITY, typed `readonly CapturedTransaction[]`.** A deploy script therefore holds the same object and can read the list mid-run. That is the accepted consequence the spec ratifies, not a feature: I added no hook, no callback, no second capture-aware entry point, and documented nothing about mid-run reads. The `readonly` element type is the cheap half-measure that stops a consumer pushing into the run's own list at the type level.

6. **The funnel that does NOT reach the choke point, confirmed in code: `recoverTransactionsIfAny`** (`packages/rocketh/src/environment/index.ts`, the `internal` handle, run before the scripts). It adopts transactions a _previous_ run broadcast, waits for their receipts and saves the resulting deployments, so those transactions never pass this run's choke point and are absent from this run's list. I deliberately did not widen the seam: they are not this run's intent, and a fork run does not save, so it has no pending-transaction file to recover from. This is the one case where the list is not the whole of what this run's node saw happen, and I stated it in the JSDoc on `capturedTransactions` so the next reader meets it at the type. Verified as the _only_ such gap: the only three send sites in any package's `src` (`eth_sendTransaction`, and `eth_sendRawTransaction` twice) all live inside the choke point and all now capture.

7. **The tests build the environment through `loadEnvironmentFromStore` rather than calling `createEnvironment` directly**, unlike the sibling seam tests. That is what lets one helper cover both a memory run and a _fork run_ (the fork environment name and its chain identity are derived there, per ADR 0014) without the test hand-rolling the naming, which would have been a second place to keep in step with the fork rules.


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria pass). Three criteria were at real risk of being satisfied only on paper; all three are genuinely met:

1. **`signability` on the INTENT arm only.** The raw arm pushes `{type, from, raw}` and the type has no `signability` there. The test does the thing that makes this load-bearing rather than incidental: it first asserts `env.addressSignability[FACTORY_DEPLOYER] === 'unsignable'`, i.e. it pins what a lookup *would* have said, so a future author who "helpfully" adds the field to both arms goes red with the reason in front of them. That is the entry a fresh-node replay must send, and it is protected.

2. **Deferred vs `ask`.** Deferral under `throw` produces `[]` (pinned), and `ask` produces exactly one entry carrying `signability: 'unsignable'` (pinned as a whole-array equality, so a second entry from a fall-through would fail). Confirmed in the control flow too: the `ask` branch pushes once and `return`s the pasted hash, so it cannot also reach a send arm.

3. **No fee/nonce contract by accident.** The key-set test feeds an input that deliberately carries `gas`, `nonce`, `maxFeePerGas` and `maxPriorityFeePerGas`, then asserts the entry's key set exactly. That is the version of this test that can actually fail.

Also checked: capture sits on the success path of each of the three arms, never at function entry; `value` passes through as the 0x quantity (`EIP1193QUANTITY`), no bigint, no `account` name; `CapturedTransaction` is exported from `@rocketh/core`'s types, so the file-sink task can name it. Decision 6's claim that the choke point is the only send path I verified independently: the only three `eth_sendTransaction` / `eth_sendRawTransaction` sites in any package's `src` are the three inside `broadcastTransaction`, and all three now capture.

Decision 2 (optional `data`) is the right call and correctly overrides the shape block quoted in the task, since `EIP1193TransactionData.data` is itself optional and the funding transfer genuinely carries none. It is flagged as touching the file-sink task, which is where it matters.

PR #131, opened 2026-08-30
feat(write-captured-transactions-to-a-file): Write the captured transac…

wighawag:
Task: `work/tasks/done/write-captured-transactions-to-a-file.md`

Done. Here is what I built.

## What landed

`rocketh -e mainnet --is-fork --write-transactions ./batch.json` now leaves a JSON file holding exactly the transactions the run broadcast, in order, written once and atomically at the end of a successful run.

**Drift check first.** I read the entry shape that actually shipped in `capture-broadcast-transactions-on-the-environment` (its `## Decisions` in the done record, plus `CapturedTransaction` in `packages/rocketh-core/src/types.ts` and the capture sites in `packages/rocketh/src/environment/index.ts`). The two open questions had landed as the task assumed: a raw pre-signed broadcast is its own `type: 'raw'` arm carrying `from` + `raw` and no `signability`, and `value` is the 0x quantity, never a bigint. One refinement the task text does not mention: `data` is OPTIONAL on the intent arm (Decision 2 of that record), so the serializer treats it as absent-able exactly like `to` and `value`. No stale premise, so I proceeded.

**Files**

- `packages/rocketh-node/src/executor/write-transactions.ts` (new): `serializeCapturedTransactions` (explicit key-by-key mapping, exhaustive `switch` over the two arms with a `never` default) and `writeCapturedTransactions` (mkdir, write to a dotted temp file in the same directory, `renameSync` over the target, unlink the temp on failure).
- `packages/rocketh-node/src/execution-params.ts` (new): `NodeExecutionParams`, core's `ExecutionParams` plus `writeTransactions?: string`. Core untouched.
- `packages/rocketh-node/src/cli-options.ts`: the `--write-transactions <file>` option, its key in `RockethCLIOptions`, `resolveWriteTransactionsPath` (refuses an empty value), and the entry in `toExecutionParams`.
- `packages/rocketh-node/src/executor/index.ts`: `loadAndExecuteDeploymentsFromFilesWithSpecificConfig` now awaits the environment and writes the file before returning it; the execute entry points take `NodeExecutionParams`.
- `packages/rocketh-node/src/cli.ts`, `src/index.ts`: type wiring and the exported type.
- `packages/rocketh-node/test/write-transactions.test.ts` (new, 20 tests, written before the production code): option surface, exact-bytes serialization, atomicity (inode changes on rewrite, no leftover temp), and lifecycle end to end through the real run assembly (broadcast order, empty run replaces a previous file, no flag means no new file, a throwing run leaves an existing file byte-identical and creates none where there was none, fork run writes like a memory run). Every assertion is inside a per-test temp directory.
- `packages/rocketh-node/test/cli-tags.test.ts`: `writeTransactions: true` added to the exhaustive parsed-shape record.
- `packages/rocketh-node/README.md`: one row in the options table.
- `.changeset/write-captured-transactions-to-a-file.md`: `@rocketh/node: minor`.

**Verification.** `pnpm build`, `pnpm typecheck`, `pnpm format:check`, `pnpm test` (1212 tests, 102 files) all green, and I smoke-tested the built bin: `--help` prints the flag with its wording, and `-e memory --write-transactions ""` prints the refusal on stderr and exits 1. `changeset status --since=main` still reports "no changesets found" only because the file is untracked at this point; it becomes visible once the runner commits.

**The asymmetry, named as the task asked:** the flag is not gated on the run mode. On a run against a *real* network a mid-run throw means transactions really were sent and no file is written; recovery there is what rocketh's pending-transaction records are for. That is stated in the comment at the write site so the documentation task can state which run modes this output is designed for.

## Decisions

1. **The flag is `--write-transactions <file>`.** It names the WRITE, which is the only thing it controls: capture is unconditional and flagless (`env.capturedTransactions` exists on every run), so `--capture-transactions` would assert a capability the flag does not own and would send a plugin author hunting for a flag to get the in-process list. Alternatives considered: `--capture-transactions` (rejected, above), and `--write-captured-transactions` (rejected as longer without disambiguating: at the CLI there is no other set of transactions a run could write, and _captured_ is the term the in-process/glossary layer uses). This is the same discipline `CONTEXT.md` and ADR 0014 pin for `--is-fork` versus `--fork`. It touches `document-captured-transactions`, which will document the flag and its `--help` wording as shipped.

2. **The option is a `@rocketh/node` parameter, NOT a core one.** The task allowed either and called the Node-local route preferable, and it is available: `resolveExecutionParams` maps core's run parameters field by field with no spread, so an output path would have had to be added to BOTH `ExecutionParams` and `ResolvedExecutionParams` to survive the trip, widening a type every package depends on for a field only the Node runtime could ever read (and that a browser-capable core must never act on, ADR 0002). Instead `@rocketh/node` declares `NodeExecutionParams = ExecutionParams & {writeTransactions?: string}` and the two EXECUTE entry points take it; the environment-only loaders keep core's type, since a run that executes nothing has nothing to write. Core and `rocketh` are unchanged, so the ask-first change `AGENTS.md` warns about did not have to happen. It touches hardhat-deploy only as a widening (an `ExecutionParams` still passes).

3. **The writer sits INSIDE the Node run assembly, after the executor returns.** Specifically in `loadAndExecuteDeploymentsFromFilesWithSpecificConfig`, the single function both public execute entry points funnel through, which is where the environment (and therefore `capturedTransactions`) first exists on the Node side. Not the bin script: `cli.ts` ends with a bare, un-awaited `loadAndExecuteDeploymentsFromFiles(executionParams)` and never holds the returned environment. Not `_executeDeployScriptsFromFiles`, which only ever sees the RESOLVED core parameters. The placement is what makes _only on success_ free: a throw from the scripts propagates past that line, so nothing is written and any previous file is untouched.

4. **A successful run that broadcast NOTHING writes an EMPTY list, `[]`, replacing whatever was at the path.** rocketh is idempotent, so a repeat rehearsal that captures nothing is the common case rather than a corner, and leaving the previous file in place would let an operator hand a Safe yesterday's batch believing it is today's. The alternative (write nothing when the list is empty) makes _no file_ ambiguous between "the run wrote nothing" and "the run failed", which is exactly the distinction the lifecycle exists to keep sharp.

5. **Atomicity is write-then-rename within the target's own directory**, into a dotted temp name (`.batch.json.<uuid>.tmp`) so a consumer watching the directory cannot pick it up as a batch, and the temp file is removed if the write throws. Same directory because `rename` is only atomic within one filesystem. The consequence, stated in the JSDoc and deliberately kept OUT of the `--help` text: the write REPLACES the path, so pointing it at a FIFO is not a supported way to stream a run. A test observes the atomicity rather than describing it (the inode changes on rewrite, which an in-place `writeFileSync` would not do).

6. **The file's top level is a bare JSON array**, two-space indented, with a trailing newline. No envelope, no version field, no run metadata: the file has one job and the whole of it is the ordered list, which a Safe tool or a Solidity `setUp()` can walk directly. An envelope would be a new concept nobody asked for and would make the empty case (`[]`) less obviously "this run sent nothing". This is a user-visible format, so it is a hard-to-reverse default worth naming here.

7. **Each entry is serialized key by key, not passed through.** A field that ever lands on a captured entry (a gas price, a nonce, a hash) therefore cannot reach the file by accident and become a contract with consumers who would replay it. The trade-off accepted: a future legitimate field must be added here deliberately. The `switch` is exhaustive over the two arms with a `never` default, so a third arm fails to compile rather than being silently dropped, mirroring the choke point's signer switch.

8. **An empty `--write-transactions` value is REFUSED, not read as "no file".** This is a new refusal, and it is the opposite of how `--tags ''` is read, deliberately: an empty tag filter has a true meaning (select everything), while an empty path names nothing a file could be written to, and silently skipping the write ends a successful rehearsal with no batch and nothing on screen to say why, which an operator cannot detect. It throws from `resolveWriteTransactionsPath` exactly as `resolveUnknownSignerPolicy` does, so `cli.ts` turns it into a message on stderr and exit 1, BEFORE the run rather than after a rehearsal has completed. It touches the CLI's option surface only.

9. **The writer creates the parent directory when missing**, so `--write-transactions out/batch.json` is the user's answer rather than an ENOENT discovered after the whole rehearsal has run. It mirrors what the FS deployment store already does for a deployments folder.

10. **I added one row to `packages/rocketh-node/README.md`'s options table.** That table is this package's reference for the option surface, so shipping a flag without it is drift on day one. I deliberately added no prose section: the user-facing page, the glossary entry and the `--help` documentation belong to `document-captured-transactions`, which is the task that owns them.


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria pass). Three things carried real risk of passing only on paper:

1. **The flag names the WRITE.** `--write-transactions <file>`, with the reasoning recorded at the option and `--help` reading "write the transactions this run broadcast, in order, as JSON to <file> (only when the run succeeds)". It states WHEN without suggesting the path can be a stream, which matters because the write is a rename over the path. Decision 1 also rules out `--write-captured-transactions` on the right grounds. Same discipline as `--is-fork`.

2. **All-or-nothing, and the empty-list case specifically.** The subtle criterion here is not "a failed run writes nothing" (which falls out of placement) but "a successful run that broadcast nothing REPLACES yesterday's batch with `[]`". Both are pinned end to end through the real CLI: an existing file is byte-identical after a throw, no file is created where there was none, and an idempotent re-run overwrites a stale batch with `[]`. Getting that last one wrong is how an operator hands a Safe yesterday's plan, and it is covered.

3. **Atomicity and the no-accidental-fee-contract rule.** The atomicity test observes the inode changing across a rewrite rather than asserting the code says "rename", which is the version that can actually fail. And `toFileEntry` builds each entry key by key with a `never`-exhaustive switch, pinned by a test that feeds deliberately contaminated entries carrying gas/gasPrice/maxFeePerGas/nonce/hash/receipt and asserts none of those strings reach the file.

Decision 2 is the notable one and it is the right call: the option is a `@rocketh/node` parameter (`NodeExecutionParams = ExecutionParams & {writeTransactions?: string}`), so `@rocketh/core` and `rocketh` are untouched. That is the Node-local route the task called preferable, and it avoids widening a type every package depends on for a field only one runtime could read (ADR 0002).

Decision 8 (an empty `--write-transactions` value is refused rather than read as "no file") is a new refusal that diverges from how `--tags ''` is read. I am satisfied it is correct and not scope creep: it throws at the boundary before the run, so it cannot cost a completed rehearsal, and the failure it prevents is the one an operator cannot detect.

PR #132, opened 2026-08-30
feat(document-captured-transactions): Document captured transactions: w…

wighawag:
Task: `work/tasks/done/document-captured-transactions.md`

All checks are green. Summary of what I did.

## What I built

**New page: `documentation/captured-transactions/index.md`** (sidebar entry added in `.vitepress/config.mts` under _Going Further_, right after "Rehearsing on a Fork", plus a bullet in `documentation/index.md`). It covers, in this order: what the list is (the shipped `CapturedTransaction` union), what an entry deliberately omits and why, getting it in process (`env.capturedTransactions`) and as a file (`--write-transactions <file>`, with the shipped `--help` wording verbatim and the real file shape), ordering as the only promise plus rocketh's refusal to segment written as a boundary, a worked segmentation example with two Safes (so both boundary kinds appear: a change in `signability` and a change in `from`), why an intent rather than a signed transaction, the Solidity replay story with the four-term condition on address equality, the runner split for a `raw` entry, what a fixture run must turn on, and the failure/incompleteness lifecycle.

**Pointers added:** `documentation/fork-runs/index.md` gains a "Taking the rehearsal to your Safe" section plus a See-also bullet; `documentation/unknown-signers/index.md` gains a paragraph framing this as the alternative to deferring each privileged call one at a time.

**`CONTEXT.md`** gains a `captured transaction` glossary entry distinguishing it from the deployment record and from the pending-transaction record (written before the wait, deleted on confirm, for recovery), and stating it has no authority.

**`docs/adr/0012`** gains a dated `## Correction (2026-08-30)` in the shape ADR 0006 and 0012 already use, plus the superseding-spec clause. Empty changeset at `.changeset/document-captured-transactions.md`.

**Gate:** `pnpm format:check`, `pnpm typecheck`, `pnpm build`, `pnpm test` (102 files / 1212 tests) and `pnpm docs:build` all pass. Working tree carries only the intended files.

## Where the task and the code disagreed (code wins)

- The task's type block types `data` as **required** on the intent arm. Shipped, it is **optional** (`data?`), because the deterministic-factory funding transfer carries none and `EIP1193TransactionData.data` is optional. Documented as shipped, with the reason absent-stays-absent matters (a `'0x'` would replay a transfer as an empty call).
- The blocked-by done record's prose says the raw arm is `kind: 'raw'`; the shipped type uses the `type` discriminant on both arms (`{type: 'raw'; from; raw}`, no `signability`). I documented the shipped spelling. This was the "open at tasking time" question: it landed as a second arm captured verbatim, with `signability` deliberately absent.
- The task's claim about `hardhat test solidity` I re-verified rather than repeated: in the installed `@nomicfoundation/edr@0.15.0` native binary, `broadcastRawTransactionCall` is implemented via `foundry_cheatcodes::unsupported` (same set as `getWallets`, `startBroadcast`, `stopBroadcast`, `getFoundryVersion`), while `prank`, `etch` and `loadAllocs` route to real implementations. The claim holds and the page states it with the version named.
- One thing the task did not mention and the shipped JSDoc does: transactions adopted from a previous run through pending-transaction recovery never pass this run's broadcast point, so they are not in this run's list. I stated it as a one-line gap.
- Process note, not drift in the code: the task body is at `work/tasks/backlog/document-captured-transactions.md`, not `work/tasks/ready/`. I did not move it.

## Decisions

1. **The page is its own section, `documentation/captured-transactions/index.md`, not a section inside `fork-runs` or `unknown-signers`.** The feature has _two_ audiences with different run modes (a fork rehearsal producing a Safe batch, a memory run producing a Solidity fixture), and neither host page can own both: a section under `fork-runs` would imply capture is a fork feature, which is precisely what the build refused to make it, and a section under `unknown-signers` would tie it to a policy it is orthogonal to. Alternatives considered: folding the operator half into `fork-runs` and the fixture half into `testing/` (rejected as splitting one concept across two pages, so the ordering promise and its condition would have to be stated twice and could drift apart), and a subsection of `production-hardening` (rejected: this is a workflow, not a check). Sidebar text is _Captured Transactions_, matching the term now pinned in `CONTEXT.md`, so the page, the field (`env.capturedTransactions`), the glossary entry and the type name all read as one concept. It touches `.vitepress/config.mts` and the index page's "as you need them" list.

2. **The flag is documented as `rocketh`-CLI-only, with one sentence saying `hardhat deploy` does not take it today.** Verified in `packages/hardhat-deploy/src/tasks/deploy.ts`: it calls `loadAndExecuteDeploymentsFromFiles` and declares no such task option. Saying nothing would have left a hardhat-deploy reader hunting for a flag that does not exist; the sentence is scoped to today rather than promising it never will, since the file-sink task recorded that omission as an explicit non-goal rather than a refusal. It touches whoever later adds the task option: they should delete that sentence.

3. **The ADR 0012 correction says the argument overstates a design constraint as an impossibility, and names exactly which shipped type made it weaker, at which layer.** I checked the claim rather than restating the task's wording: what shipped is `SkippedExecution` in `packages/rocketh-read-execute/src/index.ts` (ADR 0013), a guarded `execute` result carrying an evaluation and no receipt. That is at the _execute_ layer, not at the seam's own return contract, which is still a hash for `savePendingExecution` / `waitForTransaction`. So the correction says the fabrication is a consequence of _where the seam sits_, a constraint that would have to be paid for rather than an impossibility, and explicitly leaves the decision standing on its load-bearing half (the capability-conditional argument, plus a deploy still owing dependent code an address). I deliberately did not soften the ADR's own text or restructure it: the trailing dated section is the form ADR 0006 and 0012 already use. The dropped-spec clause is folded into the same section and names `captured-transactions` as the superseding spec, plus _why_ it satisfies the write-only property structurally (deferred transactions produce no entry, nothing reads it back).

4. **The `CONTEXT.md` glossary entry uses a colon rather than the em-dash separator every sibling entry uses.** The repo's standing prose rule forbids em dashes, and the glossary's `- **term** — definition` shape is the one place in that file where following the rule makes an entry look unlike its neighbours. I chose `- **captured transaction** (`CapturedTransaction`, `@rocketh/core/types`): ...`, which keeps the term first and the type reference adjacent. Reversible in one edit if the maintainer prefers the file's internal consistency over the prose rule.

5. **The Solidity replay recipe is documented per runner, and the hardhat alternative is `vm.etch` at the canonical factory address with its runtime code quoted in full.** The runtime bytes are derived from the `signedTx` shipped in `packages/rocketh/src/environment/chains.ts` (the 69-byte runtime inside that initcode), not from memory, so the snippet matches the factory rocketh actually relays. I also said the preceding funding transfer can be skipped when you etch instead, because it exists only to pay for the relay. It touches nothing else, but it dates: if the default `deterministicDeployment` ever changes, that snippet must change with it.


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria pass). Three criteria carried real risk of being satisfied only on paper:

1. **The raw-replay recipe is runner-differentiated, and the EDR claim is true.** The page gives `vm.broadcastRawTransaction` for forge and explicitly refuses it for `hardhat test solidity`, routing that reader to `vm.etch` / `vm.loadAllocs` instead. I verified this against the shipped binary rather than taking the page's word: in `@nomicfoundation/edr@0.15.0` the mangled symbol `foundry_cheatcodes::unsupported::…Vm::broadcastRawTransactionCall::Cheatcode::apply` puts it in the unsupported set, while `etch` appears in the supported Vm interface. A single recipe here would have failed a hardhat reader on the FIRST entry of a fresh-chain list, and the page says exactly that. It also catches the follow-on nobody asked for: when you `vm.etch` the factory, the funding transfer that preceded the raw entry has nothing left to do.

2. **The address guarantee is conditional, with the capture-side term included.** All four terms are stated, and term 1 (the CAPTURE run must itself have started from genesis) is the one a page written from the headline drops. Without it a reader takes a mainnet-fork batch for a fixture.

3. **A successful run is not automatically a complete plan.** The failure lifecycle is documented, and then immediately qualified: a transaction deferred under `throw` produces no entry, `catchUnknownSigner` swallows the error, the run SUCCEEDS, and the file is written missing exactly the privileged calls the operator still owes. That is the trap the fork-runs page routes readers into when it tells them to turn impersonation off, and this page names it rather than letting "a failed run writes nothing" imply completeness.

Also checked: segmentation is documented as a deliberate boundary with the worked example splitting on `from` as well as `signability` (the two-Safe case); samples match the shipped `CapturedTransaction` and the shipped `--write-transactions` help text; no hook, callback, stdout mode or mid-run read is described; the ADR 0012 correction is dated and additive, leaves the decision standing, and folds in the dropped-spec pointer.

One judgement call I am accepting rather than blocking on, since the builder flagged it and it is reversible in one edit: the new `CONTEXT.md` glossary entry uses a colon separator where every sibling entry uses an em dash, choosing the repo's no-em-dash prose rule over that file's internal consistency. Correct priority, worth a maintainer's glance.

PR #133, opened 2026-08-30
feat(deferral-message-warns-about-repeat-execution): Say that a re-run…

wighawag:
Task: `work/tasks/done/deferral-message-warns-about-repeat-execution.md`

All acceptance criteria are met and the repo's checks are green (`pnpm build`, `pnpm test` 1228 passing, `pnpm typecheck`, `pnpm format:check`).

## What I did

**Drift check first.** Every premise verified against current code: `describeUnknownSignerCapabilityDegradation` is as described (`packages/rocketh/src/environment/unknownSignerPolicy.ts`); the executor reaches `internal.recordMigration` only inside `if (result && typeof result === 'boolean')` after the script function returns, while a throw is rethrown one branch earlier in the `catch` around `deployScript.func(...)` (`packages/rocketh/src/executor/index.ts:838-853` — ADR 0012 cites the old line numbers 552-557, the mechanism is unchanged); and nothing on the pasted-hash path applies a recency or block-height condition (`interactiveUnknownSigner.ts`, `classifyPastedTransaction`, `waitForPastedTransaction`, `requireDeployedContract` — the only `eth_blockNumber` read in the package is the confirmations check in `waitForTransactionReceipt`, which an older transaction passes trivially). No drift; the backlog task `document-migrations-and-run-at-the-end` has not landed, so there was no wording to match.

**Sentence one** — new pure `describeDeferralRepeatExecution(behaviour, {scopedPolicy})` next to the degradation note, same shape (returns the note or `undefined`). It says the run STOPPED before the script finished so nothing recorded that the step was reached, that this holds even for an `id` plus `return true` script because the migration is recorded only when the script RETURNS, that the next run prints the SAME transaction, that executing it twice is a real loss for a mint/transfer/increment/governance action, and then hands the reader to the remedy. It never mentions a guard. Composed into the throw branch of `broadcastTransaction` alongside the degradation note.

**Sentence two** — `formatInteractivePresentation` now states that a hash from a transaction already executed after an earlier run is accepted, and why (no freshness check).

**Tests.** `unknownSignerPolicy.test.ts`: the note's content and all four quiet/loud branches, plus `scopedPolicy()`. `unknown-signer-seam.test.ts`: message CONTENT at the seam (present under run-level `throw` and under `auto`-degraded-to-throw, absent under a scoped `throw`, no em dash, `not.toMatch(/guard/i)`). `interactive-unknown-signer.test.ts`: the behavioural pin — a transaction mined at block 1 while the chain head is at block 1000 is accepted and resolves the step — plus the prompt wording. `catchUnknownSigner.integration.test.ts`: the caught path prints the v1 block and nothing else.

**Docs + changeset.** `documentation/unknown-signers/index.md` gains "A hash from an earlier run is still good" and "A run that threw will hand you the same transaction again" (with an explicit note that the guard is a remedy, not the diagnosis). `.changeset/deferral-message-warns-about-repeat-execution.md` (`rocketh: patch`). `formatDeferredTransaction` untouched.

## Decisions

- **The quiet path is keyed on a SCOPED `'throw'` frame, not on any explicit `'throw'`.** `catchUnknownSigner` pushes `{policy: 'throw'}`, so at the seam a wrapper-scoped throw and a run-level `onUnknownSigner: 'throw'` have the same effective policy and mean opposite things about whether the run stops. To satisfy both "an unwrapped deferral under `throw` includes the warning" and "a `catchUnknownSigner` action does not", I added a private `scopedPolicy()` to the policy-frame stack (`UnknownSignerPolicyStack`, internal to `rocketh`; `@rocketh/core`'s `UnknownSignerPolicyFrame` is unchanged) and made the note silent only when a frame asked for `'throw'`. Alternative considered and rejected: mirror `describeUnknownSignerCapabilityDegradation` exactly and stay silent on every explicit `'throw'` — that would deny the note to the run-level `throw`, which is precisely the case ADR 0012 lists as "a real hazard ... and nothing warns about it today". Second alternative rejected: add a discriminator field to `UnknownSignerPolicyFrame` (a core type change, and the frame is meant to say what policy, not who pushed it). **What it touches:** `withUnknownSignerPolicy(env)('throw', ...)` is now ALSO quiet even though such a run does halt, because it is indistinguishable from `catchUnknownSigner` at the seam. I judged that acceptable (a caller who wrote `'throw'` around one call chose the defer workflow for it), and it is the one false negative in the rule. A scoped `'ask'`/`'auto'` degraded by the capability ceiling still gets the note.
- **The note is attached only on the policy-throw branch, not on the interactive degrade-to-defer exits** ("cannot sign", a cancelled/failed prompt, a refused unrelated-transaction confirmation). Those deliberately rethrow the message UNDEGRADED so `catchUnknownSigner` sees exactly what an unwrapped throw produces, and that is pinned by an existing test's stated intent; the user on that path has just READ the prompt, which now carries the stale-hash sentence. Alternative considered: warn there too, since the run does abort. Rejected as scope creep against the task's "keep it to the `throw` path's message and the interactive prompt". **What it touches:** `interactive-unknown-signer.test.ts` story 4.
- **The docs page names the guard as the SECOND remedy, under an explicit "neither of these is the diagnosis" paragraph.** The page already tells the reader that a guard is what closes the loop on a re-run, and omitting it would leave the new section contradicting the existing one. The error MESSAGE still says nothing about guards. **What it touches:** the existing "What closes the loop on the re-run" section and `../execute-guard/`.
- **No ADR written.** The choice above is small and reversible (one predicate in one pure function), and ADR 0012 already carries the underlying reasoning; the rationale lives in the function's JSDoc and here.


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria pass). This task was re-scoped before dispatch because its original rationale did not survive scrutiny: it justified the warning as "unless the step is guarded on chain state", which says nothing deferral-specific, since a guard is optional and any unguarded call re-executes on any re-run. The build had to land the corrected cause, and it did.

1. **The message names the real cause.** It says the run STOPPED before the script finished so nothing recorded that the step was reached, and it explicitly covers the idiomatic run-once script: "an 'id' plus a 'return true' records its migration only when the script RETURNS, and this deferral unwound the run before it could". That is the actual asymmetry, and it is the one that matters, because it means an author who did everything right is exposed on this path alone. A test pins it by name ("attributes it to the run stopping before the completion was recorded, not to a missing guard"), so the weaker framing cannot come back silently.

2. **The two sentences are one story.** The warning ends by pointing at the interactive paste as the way out, and the prompt states the reciprocal ("ALREADY EXECUTED IT, after an earlier run stopped on it? Paste that hash"). The stale-hash property is pinned behaviourally, not just as text: `accepts a hash for a transaction that landed long BEFORE this run`. That is the criterion that keeps a future "freshness check" from quietly removing the remedy.

3. **The quiet-path asymmetry is keyed correctly, and this was the subtle part.** `catchUnknownSigner` pushes a `{policy: 'throw'}` frame, so at the seam a wrapper-scoped throw and a run-level `onUnknownSigner: 'throw'` have the same EFFECTIVE policy and mean opposite things about whether the run stopped. Suppressing on any explicit `'throw'` (the naive mirror of `describeUnknownSignerCapabilityDegradation`) would have denied the note to precisely the case ADR 0012 calls out as an unwarned hazard. The build added `scopedPolicy()` to the internal frame stack to tell them apart, left `@rocketh/core`'s `UnknownSignerPolicyFrame` unchanged, and covered both directions with tests, including the degraded-`auto` case.

Also checked: `formatDeferredTransaction` in `@rocketh/unknown-signer` is untouched, so the v1 migration parity guarantee holds; the note is a pure function alongside its two siblings; the `'ask'` behaviour correctly returns `undefined`, since that run pauses rather than stops and the prompt carries the sentence instead.

One fixup I made on this branch before merging (5f493b7): a single em dash in the done record's Decisions block. The repo forbids the character anywhere in output, not only in message text. The message itself was already clean and is pinned by two dedicated no-em-dash tests.

Merging.

PR #134, opened 2026-08-30
feat(re-ask-a-not-found-pasted-hash-with-the-previous-value): Re-ask fo…

wighawag:
Task: `work/tasks/done/re-ask-a-not-found-pasted-hash-with-the-previous-value.md`

Done. All checks green: `pnpm build`, `pnpm typecheck`, `pnpm test` (102 files, 1240 tests), `pnpm format:check`.

## What I built

**Drift check first.** The task's premises all still hold: `waitForPastedTransaction` did throw on a not-found hash after `PASTED_TRANSACTION_LOOKUP_ROUNDS` (`packages/rocketh/src/environment/index.ts`), the malformed-paste re-ask exists at `MAX_HASH_PROMPT_ATTEMPTS = 3` (`interactiveUnknownSigner.ts`), `promptText` was `{type, name, message}` and optional, `@rocketh/web` ships no text ability, and ADRs 0006/0007 are unchanged. The only stale detail was the doc path (see Decisions).

- **`@rocketh/core`** — new `TextPromptRequest` type with an optional `initial`, documented as a HINT an implementation may ignore; `PromptExecutor.promptText` now takes it. Purely additive: existing implementations still typecheck.
- **`@rocketh/node`** — the request reaches `prompts` whole, which supports `initial` natively. JSDoc pins what that actually means there (greyed placeholder, submitted on an untouched enter, right-arrow pulls it into the buffer) and that Ctrl-C still reads as a cancellation, verified against the installed `prompts@2.4.2` source.
- **`@rocketh/test-utils`** — `MockTextPromptRequest` is now the request type itself, so `textRequests[i].initial` records what was pre-filled.
- **`rocketh` seam** — `waitForPastedTransaction` returns `{type: 'not-found'}` instead of throwing (a non-success receipt still throws, deliberately). The choke point wraps ask + lookup in a loop that re-asks with the previous hash as `previousAnswer`, and a shared `HashPromptBudget` is spent by both the malformed loop (inside the resolver) and the not-found re-ask (at the seam). The refusals stay before the pending file, gas tracker and deployment record; the address invariants run on the hash finally accepted.
- **Tests** — 8 new at the execution seam (pre-fill asserted via the recorded request, unchanged-value retry, bounded give-up by call COUNT not clock, alternating malformed/not-found, the three exits from the re-asked prompt, banner printed once), 2 at the deployment funnel (including one where the *re-asked* hash created no contract, so the invariant provably ran on the accepted hash), plus one each in `@rocketh/node` and `@rocketh/test-utils`.
- **Docs** — new "A hash that does not work out is asked again, not fatal" section in `documentation/unknown-signers/index.md`, replacing the sentence that said the run stops.
- **Changeset** — `.changeset/re-ask-a-not-found-pasted-hash.md`, minor for `@rocketh/core`, `@rocketh/node`, `@rocketh/test-utils`, `rocketh`.

## Decisions

**1. One SHARED budget of 3 questions per pause, not two.** The malformed-paste re-ask and the new not-found re-ask spend one `HashPromptBudget` created per unsignable transaction. Alternative considered: a separate not-found budget, rejected because two counters that each reset can be alternated for ever, which is exactly the unbounded loop the old hard failure existed to prevent. Kept at the existing `MAX_HASH_PROMPT_ATTEMPTS = 3` rather than raising it, so the malformed path's observable behaviour is unchanged. Touches: `MAX_HASH_PROMPT_ATTEMPTS` (re-meaned from "malformed attempts" to "text asks per pause", documented at the constant), and any future third re-ask, which must spend this budget too. Pinned by `cannot be looped for ever by alternating malformed and not-found answers`.

**2. Exhausting the budget DEFERS, it does not fail.** A not-found hash used to throw a plain `Error` naming the hash; now the run shows that message and, when the budget is gone, throws the undegraded `UnknownSignerError` exactly as "cannot sign" does. Alternative considered: keep the bespoke terminal `Error` for the not-found spender. Rejected because with a shared budget the outcome would depend on which kind of answer happened to spend the last unit, and because a plain `Error` escapes `catchUnknownSigner` and kills a run that the deferral workflow could have handled. Consequence, user-visible: a wrapper that previously died on a not-found hash now catches it and continues. Touches: `catchUnknownSigner` / `@rocketh/unknown-signer` consumers; covered by the changeset.

**3. The resolver's `cannot-sign` reason `no-valid-hash` was renamed `attempts-exhausted`.** Internal (not exported from the package, only used in a `logger.debug`), but the old name says "malformed", which becomes a lie once a never-found hash can spend the same budget.

**4. `previousAnswer` also suppresses the banner.** Its presence means "this is a re-ask within the same pause", so `formatInteractivePresentation` is printed once per pause rather than once per question, matching what the malformed loop already did. Alternative: a separate boolean flag, rejected as a second knob for one state. Documented at the parameter and pinned by `does not reprint the transaction banner on the re-ask`.

**5. Documentation went to `documentation/unknown-signers/index.md`, not `documentation.md`.** The task names a `documentation.md` that no longer exists: the docs are now the `documentation/` site (per `work/notes/ideas/docs-site-astro-starlight.md`), and the interactive-flow text the task describes verbatim lives on the unknown-signers page. Small and self-contained, so I resolved it rather than stopping, but recording it since the task text is stale.

**New named concept:** `HashPromptBudget` / `createHashPromptBudget` in `interactiveUnknownSigner.ts`. Checked against `CONTEXT.md` and the ADRs: it duplicates nothing (no existing "budget", "attempts" or "quota" concept), sits at the prompt-loop layer where both re-asks can reach it, and is module-internal, so it is not added to the glossary.

### DIFF
--- a/.changeset/re-ask-a-not-found-pasted-hash.md
+++ b/.changeset/re-ask-a-not-found-pasted-hash.md
@@ -1,0 +1,8 @@
+---
+'@rocketh/core': minor
+'@rocketh/node': minor
+'@rocketh/test-utils': minor
+'rocketh': minor
+---
+
+A transaction hash pasted at the interactive unknown-signer prompt that this node has never heard of no longer ends the run: the question is asked again with that hash offered back as the starting value, so a truncated paste, a dropped character or an RPC that had not caught up costs an edit rather than a re-run. Pressing enter on the offered value looks for the same hash again. The re-asking is bounded and shares ONE budget of three questions per pause with the existing malformed-paste re-ask, so alternating the two cannot loop; when it runs out the transaction defers exactly as `cannot sign` does, saving nothing. `PromptExecutor.promptText` requests gain an optional `initial` (the new `TextPromptRequest` type), honoured by `@rocketh/node` through the prompt library's own initial-value support, recorded by the `@rocketh/test-utils` double, and safely ignorable elsewhere; `@rocketh/web` still supplies no text ability.
--- a/documentation/unknown-signers/index.md
+++ b/documentation/unknown-signers/index.md
@@ -14,8 +14,16 @@ That is the whole workflow. There is no re-run dance, nothing to install, and an
 
 ### At the pause you have two answers
 
-- **Paste the transaction hash.** rocketh looks the transaction up on the network, waits for it to be mined, requires the receipt to report a SUCCESSFUL status, saves state through the same pending-transaction path a normal broadcast uses, records the hash for gas reporting, and returns the receipt to your script. It never sends a transaction of its own. A hash this node has never heard of (from the wrong network, or a typo that is still the right shape) is given a short grace period to show up and then reported as not found, with the transaction you still have to execute printed again, so the run stops rather than waiting for ever.
-- **`cannot sign`** (or just press enter). rocketh prints the full transaction and raises `UnknownSignerError`, which is the [defer workflow](#deferring-instead-of-asking-catchunknownsigner) below. Aborting the prompt (Ctrl-C) does the same. A paste that is not a transaction hash is re-asked a couple of times and then also defers.
+- **Paste the transaction hash.** rocketh looks the transaction up on the network, waits for it to be mined, requires the receipt to report a SUCCESSFUL status, saves state through the same pending-transaction path a normal broadcast uses, records the hash for gas reporting, and returns the receipt to your script. It never sends a transaction of its own.
+- **`cannot sign`** (or just press enter). rocketh prints the full transaction and raises `UnknownSignerError`, which is the [defer workflow](#deferring-instead-of-asking-catchunknownsigner) below. Aborting the prompt (Ctrl-C) does the same.
+
+### A hash that does not work out is asked again, not fatal
+
+A paste can be wrong in two ways, and neither ends the run. A value that is **not a transaction hash** (you pasted an address, or half a line) is rejected on the spot. A hash this node has **never heard of** (copied from the wrong network's explorer, or a character your terminal ate) is given a short grace period to show up and then reported as not found. Either way rocketh asks the same question again, and for a not-found hash it offers the value you typed BACK to you: press enter to look for that same hash once more, which is what you want when the transaction is real and the RPC had simply not caught up, or type a corrected one.
+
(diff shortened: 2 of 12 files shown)


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria pass). This task's characteristic failure is introducing an unbounded loop while removing a hard failure, so that is where I looked hardest.

1. **The bound survives, and the budget question is answered the safe way.** The two re-asks share ONE budget (`MAX_HASH_PROMPT_ATTEMPTS`, spent by both the malformed and the not-found path), rather than getting a counter each. The JSDoc states the exact hazard that rules out separate budgets: "an unfindable hash would refill whichever budget the other kind did not touch, and the run could be paused for ever". That is pinned behaviourally by `cannot be looped for ever by alternating malformed and not-found answers`, which is the one test that actually discriminates here, and termination is proven by exhausting a scripted prompt rather than by a wall clock. `attempts-exhausted` is deliberately named after the budget rather than after either spender.

2. **The pre-fill is a hint, not a contract.** `TextPromptRequest.initial` is optional and documented so an implementation with no way to offer a starting value ignores it and the caller still works; the resolver re-reads whatever was submitted rather than assuming the pre-fill came back. `@rocketh/web` is untouched and still supplies no text ability, `canPromptForText()` is unchanged (still pure method presence, ADR 0007, no TTY probe), and the `@rocketh/test-utils` double records what was pre-filled so a test can tell a genuine carry-over from the question being re-asked from scratch.

3. **Nothing is saved for a refused hash.** The pre-existing ordering holds: the refusals stay ahead of the pending-transaction file, the gas tracker and the deployment record, and the deployment path still runs its address invariants on the hash finally accepted (`runs the address invariants on the re-asked hash, and saves nothing when it created no contract`).

Also checked: all three existing exits (empty, "cannot sign", abort) still defer from the re-asked prompt; the transaction banner is not reprinted on a re-ask; the docs section that used to say a not-found hash stops the run is replaced, and correctly states the shared three-attempt bound.

A note on something I did NOT block: the diff adds ten em dashes in JSDoc. I checked before reacting, and they are consistent with this repo, whose own `AGENTS.md` carries no such rule and whose `main` already has 212 across 65 TypeScript files. Documentation prose and every user-facing string in this diff are clean, which is where this repo's tasks actually impose the rule.

Merging.

PR #135, opened 2026-08-30
feat(document-migrations-and-run-at-the-end): Document the script id /…

wighawag:
Task: `work/tasks/done/document-migrations-and-run-at-the-end.md`

## What I did

**Drift check first.** The task's premises hold in this checkout: the `id` / `return true` / `.migrations.json` mechanism is live in the executor script loop (`packages/rocketh/src/executor/index.ts`, the `hasMigrationBeenDone` skip and the `recordMigration` call after a `true` return), `runAtTheEnd` is the two-pass ordering in the same loop, `captured-transactions` has landed (page + `env.capturedTransactions` + `--write-transactions`), the deferral message ships `describeDeferralRepeatExecution`, and the unknown-signers page's interactive section is in its rewritten form. Nothing needed re-scoping. Every claim on the new page was read out of the code (executor loop, `loadDeploymentsFromStore`, `recordMigration`/`loadDeployments`, the `saveDeployments` resolution, the `_`-prefixed file filter in `@rocketh/node`), and the behaviours are the ones `packages/rocketh/test/executor-scripts.test.ts` pins (run locally, 19 passing).

**New page:** `documentation/script-lifecycle/index.md` ("Script lifecycle: running once, and running last"), covering the run-once mechanism (whole-script unit, only `true` records, the missing-`id` error, the `id` as chosen identity, applies to `runAtTheEnd` too), why it is a fast path in front of chain-derived idempotency with no partial resume, a `::: danger` warning on `return true` after a caught deferral with the `return deferred === null` fix and why rocketh documents rather than polices it (ADR 0012 amendment), the two non-hazardous paths (`ask` observed the transaction; plain `throw` records nothing AND resurfaces the same transaction, reusing the shipped wording and linking it), where `.migrations.json` lives and what wipes it, the read/write asymmetry of `hasMigrationBeenDone`, `runAtTheEnd` ordering (including that it beats the dependency graph), and the batch-proposer use case with a captured-transactions consumer plus the hand-rolled `catchUnknownSigner` collection.

**Cross-references:** sidebar (`.vitepress/config.mts`), `documentation/index.md`, `documentation/core-concepts/index.md` (next to tags/dependencies), `documentation/unknown-signers/index.md` (one sentence extending the interactive section, a `::: danger` in the `catchUnknownSigner` section, and a link from the existing run-once sentence in the throw section), `documentation/captured-transactions/index.md` (see also). Glossary entry in `CONTEXT.md`. Empty changeset per the repo convention for docs-only changes.

**Observation filed (out of scope, not fixed):** `work/notes/observations/reset-on-a-fork-run-deletes-the-forked-networks-records.md` - `--reset` on a fork run calls `deleteAll` on the forked network's folder, which a fork never writes to.

**Gate-relevant checks:** `pnpm format:check`, `pnpm sync:template:check`, `pnpm changeset status --since=main`, `pnpm build`, `pnpm typecheck` all green; `pnpm docs:build` builds the site with the new page and links. Working tree contains only the intended changes.

## Decisions

- **The page is `documentation/script-lifecycle/` ("Running Once, and Running Last" in the sidebar), not "Migrations".** The word _migration_ already means porting from hardhat-deploy v1 in this docs site (`documentation/migration/`, and the `unknown-signer-v1-migration` spec), so a page named after `.migrations.json` would make one term mean two things in the same sidebar. I kept **migration record** as the code-level name (it is what the file, `hasMigrationBeenDone` and `recordMigration` are called) and used **run-once script** as the user-facing name, which is the phrase already shipped in `documentation/unknown-signers/index.md` ("the run-once mechanism"). Alternatives considered: `documentation/migrations/` (rejected, re-means an existing term) and folding the content into `core-concepts` (rejected, the deferral hazard needs a page-level warning). This touches the CONTEXT.md glossary, where I added an entry pinning the two senses apart so the next task does not re-fork them.
- **Both features share one page rather than getting one each.** They meet in the motivating example: a `runAtTheEnd` batch proposer sits in exactly the script shape where an unconditional `return true` loses a step, so splitting them would duplicate the warning or leave one page silently incomplete. Cost: the page carries two headline concepts, mitigated by the two top-level sections and the option table at the top.
- **The `saveDeployments` sentence is stated conditionally, not as "in-memory runs do not save".** The resolution in `packages/rocketh/src/executor/index.ts` is: a fork defaults to not saving, a run with NO supplied provider defaults to saving whatever the environment is called, and only a run with a supplied provider treats `memory` / `hardhat` / `default` as non-saving. The simpler wording would have been wrong for `rocketh -e memory`, so the bullet names the provider condition.
- **Empty changeset**, following the `document-captured-transactions` precedent for a docs-only change: it touches no package, and an empty entry keeps `changeset status --since=main` honest without queueing a release. Per CONTEXT.md it must be consumed by a real changeset before publishing, which the existing non-empty entries already provide.


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria pass). This task was forward-noted before dispatch because three things landed after it was written, and the risk was a page that contradicted shipped behaviour. It did not.

1. **The `throw` path states BOTH consequences, and matches the message that already ships.** The task body called that path "safe by accident ... the re-run converges", which is only half true. The page writes both halves: nothing is recorded so no step is silently skipped, AND because nothing was recorded the re-run prints the same transaction, so an operator following the instructions twice executes a mint or a nonce-bearing governance action twice. It links the deferral-message section rather than coining a second description of the migration contract, and it does not misattribute the cause to a missing guard.

2. **Claims are verified against the executor, not against the task.** I re-checked the two most falsifiable ones myself. `recordMigration` writes `.migrations.json` immediately on the script returning, not at end of run, and is gated on `context.saveDeployments`, so the page's statement that a fork run remembers the id for that run only and writes no file is exactly right. The quoted parse-failure message matches `environment/index.ts:463` verbatim once the interpolated filename is resolved.

3. **`runAtTheEnd` leads with the shipped feature.** It reads `env.capturedTransactions` and links the captured-transactions page, rather than the dropped `deferred-transaction-collector`, and it names the boundary a reader of this page will immediately wonder about: a deferred transaction produces no captured entry.

The "rocketh does not police this, deliberately" paragraph is the one that had to be unskimmable, and it is: it gives the reason (a hand-written try/catch produces the identical outcome, so enforcement taxes the careful user and stops nobody) and cites ADR 0012's amendment rather than restating the argument.

Off-path finding correctly captured rather than fixed: `work/notes/observations/reset-on-a-fork-run-deletes-the-forked-networks-records.md`. I verified its core claim, and it holds. `loadDeployments({reset})` calls `deploymentStore.deleteAll(deploymentsFolder, environmentName)` with no fork term and no `saveDeployments` guard, so `rocketh -e mainnet --is-fork --reset` deletes `deployments/mainnet/` before a rehearsal that will never write anything back. Mitigated by the confirmation prompt naming the environment, and by committed records, but it is real and worth a follow-up task.

Merging.

PR #136, opened 2026-08-30
feat(verifier-warn-on-ambiguous-library-name): Warn when a library name…

wighawag:
Task: `work/tasks/done/verifier-warn-on-ambiguous-library-name.md`

All green. Summary of the work:

**Drift check (first, as instructed):** the premises hold. `packages/rocketh-verifier/src/library-source.ts` still resolves via `findLibrarySourcePathFromLinkReferences` first, then the `metadata.sources` scan, and both passes of that scan `return sourcePath` on the first match (AST pass, then the `\blibrary\s+<Name>\b` regex over `content`). No warning existed anywhere on that path. Proceeded.

**Changes**

- `packages/rocketh-verifier/src/library-source.ts`: split the fallback scan into `collectFromAst` / `collectFromContent`, both of which now collect ALL matching source paths instead of returning on the first. A new `chooseFirst(libraryName, candidates, pass)` returns `candidates[0]` (unchanged behaviour) and, only when `candidates.length > 1`, emits a single `console.warn` naming the library, the candidate count, every candidate path, the chosen one, and which pass produced them (AST = real duplicate declarations; text scan = a match may be a comment or string literal). The AST pass still short-circuits the text pass, so the winning candidate is byte-for-byte what it was before.
- `packages/rocketh-verifier/test/library-source-ambiguity.test.ts` (new, written before the production code and confirmed red): quiet paths (single AST candidate, single text candidate, no candidate, `linkReferences`), AST-pass ambiguity (warns once, names both paths, says which was chosen, mentions AST, does not mention comments), a guard that the text pass is not consulted once the AST pass answered, text-pass ambiguity (warns, mentions text scan + comment/string-literal, does not mention AST), and a 3-candidate case.
- `.changeset/verifier-warn-on-ambiguous-library-name.md`: `@rocketh/verifier: patch` (standing convention from `CONTEXT.md`).

**Checks:** `pnpm build`, `pnpm typecheck`, `pnpm format:check`, and the full `pnpm test` (103 files / 1248 tests) all green. Working tree contains only the three files above.

## Decisions

- **Warning channel: `console.warn`, not `named-logs`.** The task prompt says to use `named-logs` "as the rest of the repo does", but that premise does not hold for this package or this message. `@rocketh/verifier` has no `named-logs` dependency and no logger (its existing output is `console.log` + chalk), and more importantly `docs/adr/0009-user-facing-notices-stay-on-console.md` already settles this class of message: `logs()` returns a permanent no-op unless a factory was hooked FIRST, and only `packages/rocketh-node/src/cli.ts` and `hardhat-deploy` do that. The verifier's own bin (`packages/rocketh-verifier/src/cli.ts`) does not, so routing this through `named-logs` would delete, for the primary way this code runs, exactly the message the task exists to surface. It would also make the notice suppressible by `--log-level error`, which ADR 0009 explicitly rejects for user-facing notices. Alternatives considered and rejected: adding `named-logs` + `hookup()` to the verifier CLI (a user-visible logging-default change for the whole `rocketh-verify` command, far outside this task), and reusing `etherscan.ts`'s `logInfo` (chalk-yellow on stdout, but that helper is private to `etherscan.ts` and this is a stderr-worthy notice per ADR 0009). Touches: nothing else; `library-source.ts` stays dependency-free. A JSDoc at the `chooseFirst` site records the same rationale and cites ADR 0009. No new ADR: this applies an existing one rather than deciding something new.
- **Wording and shape: one `console.warn` per ambiguous resolution, prefixed `[@rocketh/verifier]`.** Format: `ambiguous library "<Name>": <n> candidate sources in the compilation metadata (<origin>): <paths…>. Using the first one: <chosen>. Etherscan keys settings.libraries by the source file that DEFINES the library, so verification will fail if that is the wrong one; a deployment carrying compiler linkReferences resolves this without guessing.` The `<origin>` clause is what distinguishes the passes: AST = "each declares it in its AST, so these are real duplicate declarations"; text = "these come from the raw-source text scan, so a match may be a mention inside a comment or a string literal rather than a declaration". The package prefix is needed because this module has no logger namespace to carry it. I deliberately did NOT de-duplicate repeated warnings across deployments (the etherscan loop calls the resolver once per linked library per deployment, so a repo-wide ambiguity can warn several times in one run): dedup requires module-level state and would suppress the per-contract context, and it is not in the acceptance criteria.
- **No new exported surface.** The candidate collectors stay module-private; `findLibrarySourcePath` / `findLibrarySourcePathFromLinkReferences` remain the only exports, and the tests drive the warning through `findLibrarySourcePath` exactly as the task's "test at the seam" instruction describes. Exporting a `findLibrarySourceCandidates` would have been a new public API on a package export surface for no acceptance-criterion gain.
- **How often the fallback is actually reached (asked for in the prompt, verified not assumed).** `packages/rocketh-deploy/src/index.ts:588` spreads the whole artifact (`...artifactToUse`) into the saved deployment, so `deployment.linkReferences` is present exactly when the user's compiler artifact carried it. The authoritative path therefore covers artifacts whose toolchain emits a populated top-level `linkReferences`. The scan is still reached in two live situations, not just "old artifacts": an artifact with an EMPTY `linkReferences: {}` (`findLibrarySourcePathFromLinkReferences` correctly returns `undefined` for it, and the repo's own fixtures at `packages/rocketh-test-utils/src/index.ts:235` and the playground artifacts are shaped that way), and libraries linked by raw name via `linkRawLibraries` when no link references exist. So the warning is not dead code, but it is bounded to contracts that both link libraries and lack usable link references. I did not change any of that; it is out of scope.


wighawag:
APPROVE ✅

Gate-3 (conductor's diff-vs-criteria pass). The stated goal was to turn a silent guess into a loud one WITHOUT changing which candidate wins, so the risks were scope creep into the resolution itself, and a warning that fires on the common case.

1. **Behaviour is unchanged; only the visibility is new.** `chooseFirst` still returns `candidates[0]` on both passes, and the AST pass still wins over the text pass (pinned by `does not consult the text scan once the AST scan has found candidates`). The `compilationTarget`-anchored resolution the task put out of scope was not attempted.

2. **The warning cannot fire on the common case, and this is pinned four ways.** Silent when the AST scan finds exactly one source, when the text scan finds exactly one, when nothing declares the library, and, most importantly, when the path came from `linkReferences` even though the sources ARE ambiguous. That last test is the one that matters, because the authoritative path involves no guessing and a warning there would train users to ignore it.

3. **The two passes are distinguished, which was the subtle requirement.** An AST-pass ambiguity says these are real duplicate declarations; a text-pass ambiguity says the match may be a mention inside a comment or a string literal. The fix a user reaches for genuinely differs between those, which is why the task asked for it. The message names the library, every candidate, the chosen one, and why it matters (Etherscan keys `settings.libraries` by the DEFINING source), and it points at `linkReferences` as the way to resolve without guessing.

I checked the AST predicate specifically, since a widened match would silently inflate the candidate set: it still requires `nodeType === 'ContractDefinition'` AND `contractKind === 'library'` AND the name, so a contract sharing a library's name is not a candidate.

On `console.warn` rather than the `named-logs` logger `AGENTS.md` generally prescribes: correct here, and for the documented reason. ADR 0009 makes a user-facing notice about a recoverable condition the explicit exception, because `named-logs` returns a no-op for every consumer that has not hooked a factory first, which includes the programmatic embedders who would hit this branch. Worth noting the surrounding verifier modules render warnings as `console.log(chalk.yellow(...))`, so this one is not yellow; I am not blocking on that, since `console.warn` is what the ADR names and it correctly goes to stderr.

Merging.

